PR #14, opened 2023-05-11
fix: fix code to wrap a value with double quotation marks

takagimeow:
I noticed that the value for the LITEFS_DIR environment variable was not enclosed in double quotes. 
To maintain consistency and avoid any potential issues, I have updated the code to ensure the value is now wrapped in double quotes. 

Please review the changes.


kentcdodds (approved):
Sure 👍 Good idea. Thank you.

PR #15, opened 2023-05-11
fix: remove extra word in email service doc

williamtstanley:
Noticed an extra word snuck in during editing...


machour (approved):

kentcdodds (approved):
Thanks!

Issue #12, opened 2023-05-11
dev command only starts server, not spinning up the ui


swalker326:
Running `npm run dev` only starts the server. 

<img width="523" alt="Screenshot 2023-05-11 at 9 39 30 AM" src="https://github.com/epicweb-dev/epic-stack/assets/8460808/71a39cc0-5845-4859-a937-80cbfd205b01">

kentcdodds:
Thanks for looking into the stack @swalker326! I'm afraid I can't reproduce this issue.

<img width="932" alt="image" src="https://github.com/epicweb-dev/epic-stack/assets/1500684/513e6d2e-1986-49d6-8be8-295674323508">

Could you dig a little more?

kentcdodds:
Ah, yeah, this project uses `npm` so I'll close this out. Happy to accept a PR to make `pnpm` work if it doesn't negatively impact the experience of anything else. Thanks!

PR #25, opened 2023-05-12
Explicitly install some dependencies that were missing

markgaze:
I was looking to use this stack with `pnpm` and spotted #12 which I could recreate locally when not using `npm`. As per [this comment](https://github.com/epicweb-dev/epic-stack/issues/12#issuecomment-1544338373), this PR installs all of the missing dependencies that makes both `pnpm dev` and `pnpm test` fail.

You can see the original issues by running `pnpm i` and then `pnpm typecheck`. The dependencies installed in this PR make `pnpm typecheck` pass correctly (as it does with `npm`).

It should have no side-effects on the `npm` usage of this stack as I've installed the same versions that were already in use but not explicitly installed via the `package.json`


kentcdodds (approved):
Thanks!

PR #31, opened 2023-05-13
Bikeshedding: use timing reporter with cachified

Xiphe:
Hey Kent,

that was a great talk at remix-conf and i'm exited that you've chosen to make cachified part of this stack 🤗

The way the timings were taken in the cachified wrapper to me felt a little like woking _around_ the library and I wanted to try if we could model this as a reporter instead.

Let me know what you think.

### DIFF
--- a/app/utils/cache.server.ts
+++ b/app/utils/cache.server.ts
@@ -157,6 +157,7 @@ export async function searchCacheKeys(search: string, limit: number) {
 export async function cachified<Value>({
 	request,
 	timings,
+	reporter = verboseReporter(),
 	...options
 }: CachifiedOptions<Value> & {
 	request?: Request
@@ -164,9 +165,6 @@ export async function cachified<Value>({
 }): Promise<Value> {
 	return baseCachified({
 		...options,
-		reporter: mergeReporters(
-			cachifiedTimingReporter(timings),
-			options.reporter ?? verboseReporter(),
-		),
+		reporter: mergeReporters(cachifiedTimingReporter(timings), reporter),
 	})
 }
--- a/app/utils/cache.server.ts
+++ b/app/utils/cache.server.ts
@@ -166,8 +166,7 @@ export async function cachified<Value>({
 		...options,
 		reporter: mergeReporters(
 			cachifiedTimingReporter(timings),
-			options.reporter,
-			verboseReporter(),
+			options.reporter ?? verboseReporter(),
 		),
 	})
 }
--- a/app/utils/cache.server.ts
+++ b/app/utils/cache.server.ts
@@ -155,12 +155,10 @@ export async function searchCacheKeys(search: string, limit: number) {
 }
 
 export async function cachified<Value>({
-	request,
 	timings,
 	reporter = verboseReporter(),
 	...options
 }: CachifiedOptions<Value> & {
-	request?: Request
 	timings?: Timings
 }): Promise<Value> {
 	return baseCachified({
--- a/app/utils/cache.server.ts
+++ b/app/utils/cache.server.ts
@@ -4,6 +4,7 @@ import {
 	cachified as baseCachified,
 	lruCacheAdapter,
 	verboseReporter,
+	mergeReporters,
 	type CacheEntry,
 	type Cache as CachifiedCache,
 	type CachifiedOptions,
@@ -13,7 +14,7 @@ import { getInstanceInfo, getInstanceInfoSync } from 'litefs-js'
 import LRU from 'lru-cache'
 import { z } from 'zod'
 import { updatePrimaryCacheValue } from '~/routes/admin+/cache_.sqlite'
-import { time, type Timings } from './timing.server'
+import { cachifiedTimingReporter, type Timings } from './timing.server'
 import { singleton } from './singleton.server'
 
 const CACHE_DATABASE_PATH = process.env.CACHE_DATABASE_PATH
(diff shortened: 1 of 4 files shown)


app/utils/cache.server.ts
```diff
@@ -161,29 +162,12 @@ export async function cachified<Value>({
 	request?: Request
 	timings?: Timings
 }): Promise<Value> {
```
Xiphe, line 164: Now that the implementation is this easy it might be more straight forward to remove the wrapper altogether in favor of something like

```ts
import { cachified } from 'cachified'
import { baseLruCacheOpts } from '~/utils/cache.server'

cachified({ ...baseLruCacheOpts, key: 'asdf', getFreshValue });
```

But thats hard to decide on before this is actually being used somewhere in the app :)

Also I was unsure why the request is passed here. I remember that on your website the cache was disabled for admins right? Do you have plans to use request in here?

app/utils/timing.server.ts
```diff
@@ -1,8 +1,25 @@
...
+
+			if (!timingType) {
+				// eslint-disable-next-line no-multi-assign
+				timingType = timings[type] = []
+			}
+			timingType.push({ desc, type, time: performance.now() - start })
+		},
+	}
+}
+
```
Xiphe, line 22: not entirely happy with this abstraction. Needed something where the start and end could be called manually. Happy to take suggestions for other directions.

package.json
```diff
@@ -74,7 +74,7 @@
     "address": "^1.2.2",
     "bcryptjs": "^2.4.3",
     "better-sqlite3": "^8.3.0",
-    "cachified": "^3.4.1",
+    "cachified": "^3.5.3",
```
Xiphe, line 77: I've added `mergeReporters` and the `"done"` event in `v3.5` _(and then had a build issue for the next 3 releases 🙈)_

kentcdodds:
I'm definitely interested in this! I'll give it a look soon.

app/utils/cache.server.ts
```diff
@@ -161,29 +162,12 @@ export async function cachified<Value>({
...
-	})
-	const result = await time(cachifiedPromise, {
-		timings,
-		type: `cache:${options.key}`,
-		desc: `${options.key} cache retrieval`,
+		reporter: mergeReporters(
+			cachifiedTimingReporter(timings),
+			options.reporter,
+			verboseReporter(),
+		),
```
kentcdodds, line 171: This makes it impossible to override (disable) the `verboseReporter` as it was before. Can we refactor it a bit to make that possible? Will this work?

```suggestion
		reporter: mergeReporters(
			cachifiedTimingReporter(timings),
			options.reporter ?? verboseReporter(),
		),
```

app/utils/timing.server.ts
kentcdodds: Looks ok to me. I can't think of a better way.

kentcdodds (commented):
Thanks! Just a bit of feedback.

app/utils/cache.server.ts
```diff
@@ -156,34 +157,14 @@ export async function searchCacheKeys(search: string, limit: number) {
 export async function cachified<Value>({
 	request,
 	timings,
+	reporter = verboseReporter(),
```
Xiphe, line 160: moved the default reporter here so it can be disabled with `null`

Xiphe:
Thanks for the feedback Kent! Made it so that the verbose reporter can either be disabled with `null` or replaced with a custom one.

app/utils/cache.server.ts
kentcdodds: Oh, I forgot your comment here. Ummm... Let's remove it.

Xiphe: Should we even create the `defaultReporter`?

with the situation that we want to be able to disable the verbose reporter it's probably a good start to go with 

```ts
import { cachified, verboseReporter, mergeReporters } from 'cachified'
import {  lruCache } from '~/utils/cache.server'
import { timingReporter } from '~/utils/timings.server'

cachified({
  cache: lruCache,
  reporter: mergeReporters(timingReporter(timings), verboseReporter()),
  key: 'asdf',
  getFreshValue
});
```

and then maybe create abstractions later when we know more about the common usages.

## Alternative 1: what I proposed above, but with option to disable verbose

```ts
import { cachified } from 'cachified'
import { 
  lruCache,
  defaultReporter // (t) => mergeReporters(timingReporter(t), verboseReporter())
} from '~/utils/cache.server'

const verbose = true;
cachified({ cache: lruCache, reporter: defaultReporter({ timings, verbose }), key: 'asdf', getFreshValue });
```

## Alternative 2: create option sets for lru and db caches

```ts
import { cachified } from 'cachified'
import {  lruCacheOpts, dbCacheOpts } from '~/utils/cache.server'

cachified({
  ...lruCacheOpts({ timings, reporter: null }),
  key: 'asdf',
  getFreshValue
});
```

kentcdodds: Sorry, I misspoke, I meant let's remove `request` but keep our little wrapper with default options.

Xiphe: Ah ok, that makes sense! The timings reporter already is a [no-op when no timings object is passed](https://github.com/epicweb-dev/epic-stack/pull/31/files#diff-9918f0627f11a6c837483de1b259a0deab76d827522275110c45e1f86d069fadR69).

Gonna remove the `request`. And ping you for a final review.

Xiphe:
removed the `request` option for the wrapper. and the timingReporter is already noop when no timing is passed. So we're good to go here from my perspective. 🚀

Have a wonderful day ☀️

kentcdodds (approved):
Thanks!

PR #33, opened 2023-05-14
test: add missing await to context.addCookies call

mxschmitt:
Hurts to create such a small PR but feels right thing to do.


kentcdodds (approved):
Thanks! Not sure how I missed this. It's appreciated!

PR #24, opened 2023-05-12
Update few forms to use conform

rajeshg:
Recently started using conform myself, so I don't expect this to be perfect. Any feedback is appreciated.

1. Added conform in some forms.
2. Added requireUserId to prevent unauthorized access to notes pages.

PR #35, opened 2023-05-14
Update profile image route using conform

lgastler:
This PR is similar to #24 but just focused on the Profile Photo route (`settings/profile/photo`) since I implemented something similar (image upload) using conform just the other day. 

I tried to stick to the approaches (regarding conform and forms) used on other routes and in #24. There seems to be a [bug](https://github.com/remix-run/web-std-io/pull/28) with `@remix-run/web-fetch` which needs some adjustments to the `zod` schema. This is documented on the conform docs [here](https://conform.guide/file-upload). 

I also tried to add a simple test to check the file size but failed to find a way to provide a file with the necessary size without adding one to the fixtures. If anyone has any ideas, I'd be happy to try again.

### DIFF
--- a/app/routes/settings+/profile.photo.tsx
+++ b/app/routes/settings+/profile.photo.tsx
@@ -1,3 +1,5 @@
+import { conform, useForm } from '@conform-to/react'
+import { getFieldsetConstraint, parse } from '@conform-to/zod'
 import * as Dialog from '@radix-ui/react-dialog'
 import {
 	type DataFunctionArgs,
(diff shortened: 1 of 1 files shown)


app/routes/settings+/profile.photo.tsx
```diff
@@ -21,17 +23,29 @@ import { authenticator, requireUserId } from '~/utils/auth.server'
...
 	Button,
+	ErrorList,
 	getFieldsFromSchema,
 	LabelButton,
-	preprocessFormData,
-	useForm,
 } from '~/utils/forms'
 import { getUserImgSrc } from '~/utils/misc'
 
 const MAX_SIZE = 1024 * 1024 * 3 // 3MB
```
rajeshg, line 32: 3 MB could be quite restrictive as a default. 5 MB may be?

rajeshg (changes requested):
Overall I like this. Thanks for making this change.

app/routes/settings+/profile.photo.tsx
lgastler: From my side gladly, but I think the whole thing will be interesting later when the whole file and image strategy solidifies. From what I've seen in videos (especially the interview with Kurt Mackey) the plan could be to outsource the whole image strategy, maybe a separate SQLite just for images, that depends on how small or large you want to keep the images and DB in general.

kentcdodds: I'm thinking it would be great to figure out a way to compress images before upload. I'm pretty sure that's possible.

kentcdodds (approved):
Looks great to me. Thanks!

PR #40, opened 2023-05-15
feat: add Sentry

scefali:
This PR adds Sentry to the Epic Stack. I am largely just following the guide Sentry has for Remix: https://docs.sentry.io/platforms/javascript/guides/remix/


README.md
```diff
@@ -211,6 +211,14 @@ Every commit to your `main` branch will trigger a deployment to your production
 environment, and every commit to your `dev` branch will trigger a deployment to
 your staging environment.
 
+- **Create a Sentry account.** (Can be deferred to later)
+
+NOTE: this step is only required if you want to use Sentry for monitoring.
+
+Create a Sentry account with a Remix project [here](https://sentry.io/signup/?project_platform=javascript-remix)
+
+Copy the DSN from the code snippet to your `fly.toml` file in the `env` section and name it `SENTRY_DSN`.
```
scefali, line 220: Not sure how detailed we want the instructions to go. Note that the flow might not be super well optimized for this template because we show code snippets/instructions that's already there:

Screen Shot 2023-05-15 at 12 54 18 PM

If we wanted, Sentry could build out a flow that will take the user directly to the DSN page as shown below:


Screen Shot 2023-05-15 at 12 55 37 PM

Just lmk :)

app/entry.client.tsx
```diff
@@ -5,6 +5,9 @@ import { hydrateRoot } from 'react-dom/client'
 if (ENV.MODE === 'development') {
 	import('~/utils/devtools').then(({ init }) => init())
 }
+if (ENV.MODE === 'production') { 
+	import('~/utils/monitoring.client').then(({ init }) => init())
```
scefali, line 9: I'm not sure if this is the best way to do this but I just copied what we do for devtools

app/routes/_marketing+/index.tsx
```diff
@@ -133,6 +133,11 @@ export default function Index() {
 								alt: 'ESLint',
 								href: 'https://eslint.org',
 							},
+							{
+								src: 'https://svgur.com/i/t9k.svg',
```
scefali, line 137: Maybe we want to host his somewhere else?

fly.toml
```diff
@@ -12,6 +12,9 @@ auto_rollback = true
 source = "data"
 destination = "/data"
 
+[env]
+# SENTRY_DSN =
```
scefali, line 16: Sentry DSNs are not sensitive so I think it make more sense to set it here rather than via CLI. But this makes it a bit different so I'm definitely open to other strategies.

priscilawebdev:
@scefali I've roughly checked this PR and have yet to get myself familiar with 'epic-stack'.  Do you mind sharing why the source maps still aren't working?

scefali:
> @scefali I've roughly checked this PR and have yet to get myself familiar with 'epic-stack'. Do you mind sharing why the source maps still aren't working?

@priscilawebdev I need to add the step to upload source maps, working on that now. Trying to figure out the right way of handling all the env variables though.

README.md
kentcdodds: We should probably move everything to the docs eventually and have only basic instructions or just a link in the README. Would you mind adding a docs page for this and linking to that in this step to get us started?

app/root.tsx
```diff
@@ -137,6 +138,7 @@ export default function App() {
 		</html>
 	)
 }
+export default withSentry(App);
```
kentcdodds, line 141: What happens here if people never set up sentry?

app/routes/_marketing+/index.tsx
kentcdodds: I just drag-and-drop the image into a github textbox and copy the URL that it gives me.

fly.toml
kentcdodds: I would rather do stuff like this in the dockerfile, but even this I think is better as part of the env file so it can be changed more easily.

kentcdodds (commented):
Cool! Thanks for the work on this so far!

fly.toml
scefali: @kentcdodds Hmm so the problem with doing it in a `.env` file is that it won't ship during a deployment. How do we get around having to duplicate your environment variables since we need it during the Fly deployments?

app/root.tsx
scefali: @kentcdodds if you don't have a DSN set, Sentry basically becomes a no-op. Everything will work as usual.

app/entry.client.tsx
scefali: @kentcdodds thoughts on this strategy?

README.md
scefali: @kentcdodds when you say "move everything to docs", what parts are you referring to?

fly.toml
kentcdodds: > having to duplicate your environment variables 

I'm not sure I follow. Why do we need to duplicate? We just set it once like we do with the SESSION_SECRET etc. and then we don't worry about it anymore.

README.md
kentcdodds: I mean we should have a docs page for the setup process instead of sticking everything in the README.md. I'll make that change today probably so it's more clear for you where this should go.

kentcdodds:
Sorry it took me so long to get around to updating the docs, but they're up-to-date now so if you could follow the same sort of thing we do for mailgun that would be stellar 👍 The doc name should be "error-reporting.md" not "sentry.md"

Thanks again!

scefali:
@kentcdodds No worries...I'm still trying to figure out the best way of getting source maps set up. Not the easiest thing! Might take me a few weeks to get everything ready.

app/routes/_marketing+/logos/logos.ts
```diff
@@ -112,4 +112,9 @@ export const logos = [
 		alt: 'ESLint',
 		href: 'https://eslint.org',
 	},
+	{
+		src: 'https://svgur.com/i/t9k.svg',
```
kentcdodds, line 116: In the spirit of making sure everything works offline, could you add this svg to the project directly?

epicstack.db
kentcdodds, line 1: I think this file needs to be deleted.

package.json
```diff
@@ -136,5 +138,8 @@
   },
   "prisma": {
     "seed": "tsx prisma/seed.ts"
+  },
+  "volta": {
+    "node": "18.16.0"
```
kentcdodds, line 143: I don't think I'm ready to commit to volta just yet. Let's remove this.

fly.toml
```diff
@@ -12,6 +12,9 @@ auto_rollback = true
 source = "data"
 destination = "/data"
 
+[deploy]
+release_command = "npm run upload:sentry"
```
kentcdodds, line 16: I think I'd rather move this command to the Dockerfile.

package.json
scefali: @kentcdodds Oh yea this is my bad, I'll make sure to remove it.

scefali:
> This all looks great to me! Just a few things. Also, could we get a docs page?

@kentcdodds Yea, I'll get to this once the underlying bug in the SDK is fixed. Unfortunately, that blocks getting this merged, right now it's not even functional because of this bug https://github.com/getsentry/sentry-javascript/issues/8265

kentcdodds:
Sounds great! Thank you 👏

fly.toml
scefali: @kentcdodds Unfortunately, I don't think this is possible. The environment variables from Fly.io are not available during build time, it has to come post-build it seems:

https://community.fly.io/t/reading-env-variables-during-deployment/12935/2

fly.toml
scefali: @kentcdodds Missed this comment earlier. But if you look at `SESSION_SECRET`, you'll see that it exits in the `.env` file and there's also a separate step for `fly secrets set`. But I think that's unavoidable.

kentcdodds: Oh, I see what you mean. Yeah, the `.env` is just generated anyway because you don't run with sentry enabled locally anyway.

app/utils/monitoring.server.ts
```diff
@@ -0,0 +1,9 @@
+import * as Sentry from '@sentry/remix'
+
+// TODO: Add Prisma integration
```
scefali, line 3: For some reason, the Prisma integration was causing deployments to fail. I didn't spend a lot of time on this but we probably can leave it out for now

scefali:
@kentcdodds Ok I think this is finally ready to go! Not a perfect implementation (couldn't get the Sentry-Prisma integration working) but at the very least it's a good starting place I think.

sergiodxa:
With Remix v1.17.0 there's now a `handleError` function in entry.server that can work for error reporting.

kentcdodds:
Main is now on the latest Remix as well :)

app/utils/monitoring.server.ts
kentcdodds: How would this change if I wanted to try the prisma integration myself?

docs/monitoring.md
```diff
@@ -0,0 +1,35 @@
...
+
+> **NOTE**: this is an optional step and only needed if you want monitoring in
+> production.
+
+## SaaS vs Self-Hosted
+
+Sentry offers both a [SaaS solution](https://sentry.io/) and
+[self-hosted solution](https://develop.sentry.dev/self-hosted/). This guide
+assumes you are using SaaS but the guide still works with self-hosted with a few
+modifications.
```
kentcdodds, line 15: I *love* this. Very much in the ethos of Epic Stack of avoiding vendor lock-in 👍 Thanks!

docs/monitoring.md
```diff
@@ -0,0 +1,35 @@
...
+becomes `SENTRY_DSN`). Then click
+[this](https://sentry.io/orgredirect/settings/:orgslug/developer-settings/new-internal/)
+to create an internal integration. Give it a name and add the scope for
+`Releases:Admin`. Press Save, find the auth token at the bottom of the page, and
+copy that to secure location (this becomes `SENTRY_AUTH_TOKEN`). Then vist the
+organization settings page and copy that organization slug (`SENTRY_ORG_SLUG`).
+
+Now, set the secrets in Fly.io:
+
+fly secrets set SENTRY_DSN=<your_dsn> SENTRY_AUTH_TOKEN=<your_auth_token> SENTRY_ORG=<your_org_slug> SENTRY_PROJECT=javascript-remix
```
kentcdodds, line 35: Is "javascript-remix" really supposed to hard coded? If so, maybe we should just put that in the Dockerfile?

````suggestion
```sh
fly secrets set SENTRY_DSN=<your_dsn> SENTRY_AUTH_TOKEN=<your_auth_token> SENTRY_ORG=<your_org_slug> SENTRY_PROJECT=javascript-remix
```
````

File: package.json (hunk 7 above)
Comment (kentcdodds) [reply package.json:136]: Reminder ☝️  :) Thanks!

File: scripts/sentry-create-release.js [hunk 12]
```diff
@@ -0,0 +1,19 @@
...
+const DEFAULT_BUILD_PATH = 'public/build'
+
+// exit with non-zero code if we have everything for Sentry
+if (
+	process.env.SENTRY_DSN &&
+	process.env.SENTRY_ORG &&
+	process.env.SENTRY_PROJECT &&
+	process.env.SENTRY_AUTH_TOKEN
+) {
+	createRelease({}, DEFAULT_URL_PREFIX, DEFAULT_BUILD_PATH)
```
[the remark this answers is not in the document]
kentcdodds: Very interesting!

server/index.ts
```diff
@@ -86,7 +94,7 @@ app.use(
 		crossOriginEmbedderPolicy: false,
 		contentSecurityPolicy: {
 			directives: {
-				'connect-src': MODE === 'development' ? ['ws:', "'self'"] : null,
+				'connect-src': MODE === 'development' ? ['ws:', "'self'"] : ["*.ingest.sentry.io"],
```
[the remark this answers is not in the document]
kentcdodds: Thanks for that! Could we make this conditional based on whether the environment variable for the Sentry DSN is set?

fly.toml
kentcdodds: Could we change this to `node ./scripts/sentry-create-release` so we don't need the npm script?

kentcdodds (changes requested):
This is so cool of you to spend time helping make this happen. I'm really excited to get this merged 👍 I just have a couple questions/suggestions.

docs/monitoring.md
scefali: @kentcdodds Yea it is supposed to be hardcoded because that's the name of the project that will get created, at least if you do it through new user onboarding (we don't ask for the project name there). Though if you change the project name then you'll have to make a change here.

docs/monitoring.md
scefali: @kentcdodds Yep! If you don't want to pay for Sentry, you can just run it yourself! Gotta love that open source mentality :D

scefali:
> How would this change if I wanted to try the prisma integration myself?

@kentcdodds Do you want me to put a comment on how to use it assuming it works correctly?

docs/monitoring.md
kentcdodds: Oh I see, so users can put whatever value they want here and then reports will be associated to this project? Should this be set differently based on environment? We could do something for this in the setup script where we determine the app name.

app/entry.client.tsx
```diff
@@ -5,6 +5,9 @@ import { hydrateRoot } from 'react-dom/client'
 if (ENV.MODE === 'development') {
 	import('~/utils/devtools.tsx').then(({ init }) => init())
 }
+if (ENV.MODE === 'production') {
```
kentcdodds, line 8: I wonder if we should add the sentry dsn env var to this check so we can have this bundle not load if they've not setup sentry yet. Would that do any harm?

app/entry.server.tsx
```diff
@@ -12,6 +12,10 @@ const ABORT_DELAY = 5000
 init()
 global.ENV = getEnv()
 
+if (ENV.MODE === 'production') {
```
kentcdodds, line 15: Same thought here 🤔

app/utils/monitoring.client.tsx
```diff
@@ -0,0 +1,31 @@
...
+					useMatches,
+				),
+			}),
+			// Replay is only available in the client
+			new Sentry.Replay(),
+		],
+
+		// Set tracesSampleRate to 1.0 to capture 100%
+		// of transactions for performance monitoring.
+		// We recommend adjusting this value in production
```
kentcdodds, line 23: Can you elaborate on this? What's the recommended value? Why don't we set this to the recommended value?

kentcdodds (commented):
Sorry for so much back and forth. Thanks for your patience!

server/index.ts
```diff
@@ -86,7 +94,12 @@ app.use(
 		crossOriginEmbedderPolicy: false,
 		contentSecurityPolicy: {
 			directives: {
-				'connect-src': MODE === 'development' ? ['ws:', "'self'"] : null,
+				'connect-src':
+					MODE === 'development'
+						? ['ws:', "'self'"]
+						: process.env.SENTRY_DSN
+						? ['*.ingest.sentry.io']
```
TryingToImprove, line 101: If self-hosted, would the URL here not be different? Some blockers (https://github.com/easylist/easylist/issues/6963) will block sentry.io, and therefor self-hosted/tunnel can be used

app/utils/monitoring.client.tsx
scefali: @kentcdodds Honestly, I just copied it from the default onboarding but yea we can change this. The right amount depends on your own needs but I think 10% is a reasonable default.

app/entry.client.tsx
scefali: @kentcdodds No harm, I can make that change.

docs/monitoring.md
scefali: @kentcdodds  The name has to match what you have in Sentry, but the default project name for a Remix project is `javascript-remix`. You don't need a different project per env as Sentry will automatically pick up the `NODE_ENV` environment variable and tag that for each event. Though it's perfectly valid to make a separate project if you want as well, people do it both ways.

server/index.ts
scefali: @TryingToImprove Yes, the self-hosted or tunnel URL would be different. I can add a comment.

kentcdodds (changes requested):
Thanks a lot! I'm happy with what we've got here. Just one last thing and we'll get this merged.

scefali:
@kentcdodds Done, just added a comment about the project name.

kentcdodds (approved):
Excellent. Thank you so much!

PR #47, opened 2023-05-16
Upgrade to Native ESM

### LINKED ISSUE #28: Upgrade to Native ESM
I'm pretty sure we want to use native ESM in this project. I believe Remix v2 will default to native ESM. It may be more trouble than it's worth, but I think it's worth looking into.

onemen:
resolve #28 

Maybe it will be easy to work on this here

Would you like to upgrade [remix.init](https://github.com/onemen/epic-stack/tree/main/remix.init) files to ESM


server/index.ts
```diff
@@ -8,7 +9,10 @@ import closeWithGrace from 'close-with-grace'
 import { createRequestHandler } from '@remix-run/express'
 import { broadcastDevReady } from '@remix-run/node'
 
-const BUILD_DIR = path.join(process.cwd(), 'build')
+const pathParts = [process.cwd(), 'build']
+if (process.env.NODE_ENV === 'production') pathParts.push('index.js')
+const BUILD_DIR = path.join(...pathParts)
+const BUILD_DIR_FILE_URL = pathToFileURL(BUILD_DIR).href
```
kentcdodds, line 15: Does pushing `index.js` on here not work in non-production environments? Based on what I'm reading below, I'm pretty sure we can rework a few things so we can just do:

```suggestion
const BUILD_DIR = path.join(process.cwd(), 'build')
const BUILD_DIR_FILE = path.join(BUILD_DIR, 'index.js')
const BUILD_DIR_FILE_URL = pathToFileURL(BUILD_DIR).href
```

And then make a few adjustments below.

In fact, I wonder whether `BUILD_DIR_FILE_URL` is necessary, couldn't we just `await import(BUILD_DIR_FILE)`?

other/test-setup/vitejs_plugin_react.cjs
```diff
@@ -0,0 +1,2 @@
+// react types are missing when import as default to ESM module
```
kentcdodds, line 1: Let's rename this file to `vitejs-plugin-react.cjs`

package.json
```diff
@@ -5,12 +5,13 @@
   "license": "MIT",
   "epic-stack": true,
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com/)",
+	"type": "module",
```
kentcdodds, line 8:
```suggestion
  "type": "module",
```

package.json
```diff
@@ -5,12 +5,13 @@
...
   "scripts": {
     "build": "run-s build:*",
     "build:remix": "remix build",
     "build:server": "tsx ./other/build-server.ts",
     "dev": "remix dev -c \"npm run dev:server\" --no-restart",
-    "dev:server": "cross-env MOCKS=true tsx watch --clear-screen=false --ignore \"app/**\" --ignore \"build/**\" --ignore \"node_modules/**\" ./index.js",
+    "dev:server": "cross-env MOCKS=true tsx watch --clear-screen=false --ignore \"app/**\" --ignore \"build/**\" --ignore \"node_modules/**\" ./index.cjs",
```
kentcdodds, line 14: Any reason we can't make the entry file an ESM file with dynamic imports for the dynamic stuff?

kentcdodds (changes requested):
Ok, I think we should proceed here. Just a few bits of feedback.

kentcdodds:
Oh, and let's figure out why playwright's failing and get that passing.

package.json
onemen: done

package.json
onemen: done

other/test-setup/vitejs_plugin_react.cjs
onemen: done

server/index.ts
onemen: We need to use `BUILD_DIR_FILE_URL` on Windows, see this error message

```shell
Error: Only URLs with a scheme in: file and data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at __node_internal_captureLargerStackTrace (node:internal/errors:490:5)
    at new NodeError (node:internal/errors:399:5)
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/resolve:1059:11)
    at defaultResolve (node:internal/modules/esm/resolve:1135:3)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at u (file:///C:/projects/kcd-apps/epic-stack/node_modules/@esbuild-kit/esm-loader/dist/index.js:1:2406)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at nextResolve (node:internal/modules/esm/loader:163:22)
    at ESMLoader.resolve (node:internal/modules/esm/loader:838:24)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:7) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```

we can use `const BUILD_DIR = '../build/index.js'`

kentcdodds:
Sorry, I just pushed a bunch of changes that impact your PR. That's my bad. I can clean it up if you don't beat me to it.

onemen:
@kentcdodds ,

Would you like to upgrade [remix.init](https://github.com/onemen/epic-stack/tree/main/remix.init) files to ESM>

kentcdodds:
Yes

kentcdodds:
Ah, yeah, why don't we make the `index.js` just do this:

```js
import('./index.mjs')
```

I'm trying to think whether there would be anything else we'd want to do to make sure errors properly exit the process if we do that.

onemen:
> Ah, yeah, why don't we make the `index.js` just do this:
> 
> ```js
> import('./index.mjs')
> ```
> 
>

we need to export the function we importing from index.mjs, if this is acceptable, i can update the PR. 

`index.js`
```js
module.exports = async (...args) => {
	const { default: main } = await import('./index.mjs')
	main(...args)
}
```

onemen:
maybe we can use `index.ts`  `@remix-run/dev/dist/cli/commands.js` have a the code to convert it to js

onemen:
@kentcdodds,

Did you read my last two comments regarding remix.init?

kentcdodds:
I like this idea:

```js
module.exports = async (...args) => {
	const { default: main } = await import('./index.mjs')
	await main(...args)
}
```

onemen:
I think this PR is ready.

don't forget to squash also commit messages when merging

PR #55, opened 2023-05-17
Typescript'ified tailwind config and watching

chessydk:
As the title says, I have 

A) converted tailwind config to typescript and 
B) set up build-watch on it. In my experience this file is updated frequently (especially in the beginning of a project) and it makes sense to rebuild automatically on changes.

For B), one could argue that the file has content related to the app, so it should be located below app folder (app/styles/) which would enable automatic rebuild on change.


kentcdodds (approved):
I didn't realize you could do this! Awesome. Thank you!

PR #56, opened 2023-05-17
DRY'ing checkbox validation

chessydk:
Created new file: zodExtensions with zod schema definition for checkbox validation. Made it a function so it can set up a text-message for when the checkbox is required.


kentcdodds (changes requested):
Excellent. Can you change the filename to kebab-case instead of camelCase? Then I'll merge. Thanks!

kentcdodds (approved):
Stellar. Thank you!

PR #64, opened 2023-05-17
Initial support for client hints

kentcdodds:
This implements the ideas described in https://github.com/epicweb-dev/epic-stack/blob/main/decisions/05-client-pref-cookies.md

Eventually we'll want locale stuff in here as well I expect. For now, the timezone offset and theme preference are necessary. I would love ideas and suggestions here. I'm going to be adding full theme support (via user preference) based on this before we merge this. In that world, the client hint for the desired theme is only used when the user chooses "system".

### DIFF
--- a/app/utils/client-hints.tsx
+++ b/app/utils/client-hints.tsx
@@ -66,9 +66,10 @@ export function getHints(request?: Request) {
  * if they are not set then reloads the page if any cookie was set to an
  * inaccurate value.
  */
-export function ClientHintCheck() {
+export function ClientHintCheck({ nonce }: { nonce: string }) {
 	return (
 		<script
+			nonce={nonce}
 			dangerouslySetInnerHTML={{
 				__html: `
 const cookies = document.cookie.split(';').map(c => c.trim()).reduce((acc, cur) => {
--- a/app/utils/client-hints.tsx
+++ b/app/utils/client-hints.tsx
@@ -1,0 +1,107 @@
+import { type Request } from '@remix-run/node'
+
+const colorSchemeHint = {
+	name: 'CH-prefers-color-scheme',
+	getValueCode: `window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'`,
+	fallback: 'light',
+	transform(value: any) {
+		return value === 'dark' ? 'dark' : 'light'
+	},
+} as const
+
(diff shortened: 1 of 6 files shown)


app/utils/client-hints.tsx
```diff
@@ -0,0 +1,107 @@
+import { type Request } from '@remix-run/node'
+
+const colorSchemeHint = {
+	name: 'CH-prefers-color-scheme',
+	getValueCode: `window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'`,
```
imcodingideas, line 5: 🤯, matchMedia is new for me.

app/utils/client-hints.tsx
```diff
@@ -0,0 +1,107 @@
...
+	fallback: 'light',
+	transform(value: any) {
+		return value === 'dark' ? 'dark' : 'light'
+	},
+} as const
+
+const clockOffsetHint = {
```
smerchek, line 12: I'm not sure clock offset alone is sufficient for rendering dates on the server. One may need to render a localized date/time, including the timezone. e.g. `8:00 AM PDT` - for this, we would need to know the user's IANA timezone (which would also tell us the appropriate clock offset).

The user's timezone is a little more resilient in general to handle transitions between daylight saving time as well.

kentcdodds: Thanks! That's fair. I'm still deciding whether I need to include this in the template. I'm pretty sure all the template needs to have built-in is the color preference for dark mode. The timezone stuff can probably just be docs.

smerchek: Ya, the docs are helpful either way to be clear how folks can extend to add their own app-specific preferences. 👍

kentcdodds: I would certainly be grateful for a PR for docs like this once this is merged :)

andrecasal:
Super awesome feature Kent, thanks! I'm already using it on my platform 🎉

However, even thought the e2e tests pass, they are visually broken as the Tailwind class CSS is not being injected onto the page:
Screenshot 2023-06-03 at 12 11 06

kentcdodds:
Ah! Interesting! Could you open an issue or PR? I'll give it a look next time I get the chance.

andrecasal:
Will do.

PR #66, opened 2023-05-18
Update root.tsx

SomiDivian:
Fix `LinksFunction` when updating to remix-run v 1.16.1


kentcdodds:
Hi @SomiDivian, thanks for working on this PR. Can you tell me what problem is being solved here? I don't think this is necessary.

PR #67, opened 2023-05-18
Add ESLint, Prettier, Prisma, and Tailwind

andrecasal:
Add recommended extensions for VS Code

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+	"recommendations": [
+		"dbaeumer.vscode-eslint",
+		"esbenp.prettier-vscode",
+		"prisma.prisma",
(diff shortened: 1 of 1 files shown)


onemen:
I think that we also need a way to share `ESlint` and `Prettier` configuration to prevent VScode from automatically change the code when our local VScode configuration does not match @kentcdodds configuration

andrecasal:
At the moment, Prettier's config lives in `package.json`. Wouldn't it be more obvious where prettier's config lives if we had a `.prettierrc` file instead (that's how I overrode some of the rules in package.json btw)?

kentcdodds (approved):
Thanks!

kentcdodds:
I'm fine if someone wants to move the prettier config out of the package.json, but this isn't required for getting this merged.

PR #73, opened 2023-05-20
Disable "global" in vitest

### LINKED ISSUE #9: Disable "global" in vitest
I tried and ran into TS issues. Would love to ditch it though.

onemen:
closes #9

### DIFF
--- a/other/test-setup/setup-test-env.ts
+++ b/other/test-setup/setup-test-env.ts
@@ -1,4 +1,5 @@
 import './setup-env-vars.ts'
+import { afterAll, afterEach, expect } from 'vitest'
 import { installGlobals } from '@remix-run/node'
 import { matchers, type TestingLibraryMatchers } from './matchers.cjs'
 import 'dotenv/config'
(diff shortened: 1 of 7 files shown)


kentcdodds (commented):
Thank you! Can you check that the expect API can still use the testing library assertions with this?

other/test-setup/setup-test-env.ts
```diff
@@ -1,22 +1,12 @@
...
 
-declare global {
-	namespace Vi {
-		interface JestAssertion<T = any>
-			extends jest.Matchers<void, T>,
-				TestingLibraryMatchers<T, void> {}
-	}
-}
-
-expect.extend(matchers)
```
kentcdodds, line 18: I think we do still need to extend expect somehow so these matchers are available at runtime.

onemen: vitest

https://vitest.dev/api/expect.html#expect-extend

onemen: Yes

we need `expect.extend(matchers)` back

onemen:
now everything work

kentcdodds (approved):
Stellar! Thank you!

PR #76, opened 2023-05-20
use eslint overrides for test files

onemen:


.eslintrc.cjs
```diff
@@ -6,6 +6,15 @@ module.exports = {
 		'@remix-run/eslint-config/jest-testing-library',
 		'prettier',
 	],
+	overrides: [
+		{
+			files: ['*.test.ts', '*.test.tsx', '*.test.js', '*.test.jsx'],
+			rules: {
+				'testing-library/prefer-screen-queries': 'off',
+				'testing-library/no-await-sync-query': 'off',
```
kentcdodds, line 14: We only want these disabled for playwright tests. The vitest ones should still have this enabled.

In fact, if it's possible, we only want the testing library rules running on vitest test files. Right now they run on everything.

onemen: If we use this

```diff
module.exports = {
	extends: [
		'@remix-run/eslint-config',
		'@remix-run/eslint-config/node',
-               '@remix-run/eslint-config/jest-testing-library',
		'prettier',
	],
	overrides: [
		{
-                       files: ['*.test.ts', '*.test.tsx', '*.test.js', '*.test.jsx'],
+			extends: ['@remix-run/eslint-config/jest-testing-library'],
+			files: ['**/__tests__/**/*', '**/*.{spec,test}.*'],
+			excludedFiles: 'tests/**/*',
			rules: {
-				'testing-library/prefer-screen-queries': 'off',
-				'testing-library/no-await-sync-query': 'off',
+				'testing-library/no-await-sync-events': 'off',
+				'jest-dom/prefer-in-document': 'off',
			},
		},
	],
	rules: {
		'@typescript-eslint/consistent-type-imports': [
			'warn',
			{
				prefer: 'type-imports',
				disallowTypeAnnotations: true,
				fixStyle: 'inline-type-imports',
			},
		],
-		'testing-library/no-await-sync-events': 'off',
-		'jest-dom/prefer-in-document': 'off',
		'@typescript-eslint/no-duplicate-imports': 'warn',
	},
```

it will apply `@remix-run/eslint-config/jest-testing-library` rules only to test files that are not in `tests` folder.

however, if you want to apply `jest` and `jest-dom` rules also to playwright test then we need different override, since `@remix-run/eslint-config/jest-testing-library` combined "jest", "jest-dom", and "testing-library" plugins into one config file

kentcdodds: I only want the jest and jest-dom in the app directory so that works out great!

kentcdodds (approved):
This is perfect. Thank you!

PR #77, opened 2023-05-20
fix: pass only relavent props to each field

edmundhung:
Looks like we were passing the field props directly to the field component which end up setting extra props like `initialError`, `errorId` or `descriptionId` on the element as well. 

```html
<input id="edit-profile-name" placeholder=" " name="name" initialerror="[object Object]" form="edit-profile" errorid="edit-profile-name-error" descriptionid="edit-profile-name-description" class="h-14 w-full rounded-lg border border-night-400 bg-night-700 px-4 pt-4 text-body-xs caret-white outline-none focus:border-accent-purple disabled:bg-night-400" value="Kody">
```

This PR restrict what we are passing by using the [conform](https://conform.guide/api/react#conform) helper.

### DIFF
--- a/app/routes/_auth+/onboarding.tsx
+++ b/app/routes/_auth+/onboarding.tsx
@@ -1,4 +1,4 @@
-import { useForm } from '@conform-to/react'
+import { conform, useForm } from '@conform-to/react'
 import { getFieldsetConstraint, parse } from '@conform-to/zod'
 import {
 	json,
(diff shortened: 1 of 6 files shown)


app/routes/_auth+/onboarding.tsx
```diff
@@ -160,27 +160,25 @@ export default function OnboardingPage() {
 					<Field
 						labelProps={{ htmlFor: fields.username.id, children: 'Username' }}
 						inputProps={{
-							...fields.username,
+							...conform.input(fields.username),
 							autoComplete: 'username',
-							autoFocus: true,
+							autoFocus: typeof actionData === 'undefined' || typeof fields.username.initialError !== 'undefined',
```
edmundhung, line 165: The conform helper will set the `autoFocus` attribute for us if there is any error in the previous submission. This overrides it so that we can have the field autofocus by default, but not if there is a previous submission with no error on the username field.

app/routes/resources+/note-editor.tsx
```diff
@@ -89,6 +89,10 @@ export function NoteEditor({
 		onValidate({ formData }) {
 			return parse(formData, { schema: NoteEditorSchema })
 		},
+		defaultValue: {
+			title: note?.title,
+			content: note?.content,
+		},
```
edmundhung, line 95: Setting the default value here will allows Conform to persists the input value if there is a previous submission.

kentcdodds (approved):
Brilliant. Thank you for following up on this!

app/routes/settings+/profile.tsx
```diff
@@ -146,6 +146,10 @@ export default function EditUserProfile() {
 		onValidate({ formData }) {
 			return parse(formData, { schema: ProfileFormSchema })
 		},
+		defaultValue: {
+			name: data.user.name ?? '',
```
andrecasal, line 150: We should add `username: data.user.username,`

andrecasal: And use conform.input() in the username field too :)

andrecasal:
I'd be happy to create a PR for this @edmundhung

edmundhung:
Hi @andrecasal, thanks for checking

- I would leave `redirectTo` as is since there is no validation needed.
- It looks like Kent has fixed the username setup already on the latest main branch
- It is indeed missed. Feel free to create a PR. Thanks!

andrecasal:
Turns out Kent had already fixed both 1. and 2.

[Here's](https://github.com/epicweb-dev/epic-stack/pull/100) a PR for the app/routes/_auth+/signup.tsx.

Issue #39, opened 2023-05-15
Add nonce support


kentcdodds:
Probably using https://npm.im/helmet like I do in [my own website](https://github.com/kentcdodds/kentcdodds.com).

paolapog:
Hi Kent, could I try to help you on this?

PR #78, opened 2023-05-21
Add nonce support

paolapog:
This pr is linked to [#39](https://github.com/epicweb-dev/epic-stack/issues/39)
For this pr I've used helmet and partially re-used what was done in kentcdodds.com project.
I've used also CSP Evaluator extension (Chrome's one) to check the current CSP.

### DIFF
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -1,25 +1,31 @@
 import 'dotenv/config'
-import { PassThrough } from 'stream'
-import { type EntryContext, Response } from '@remix-run/node'
+import { PassThrough, Transform } from 'stream'
+import { Response, type HandleDocumentRequestFunction } from '@remix-run/node'
 import { RemixServer } from '@remix-run/react'
 import isbot from 'isbot'
 import { renderToPipeableStream } from 'react-dom/server'
 import { init, getEnv } from './utils/env.server.ts'
 import { getInstanceInfo } from 'litefs-js'
+import { NonceProvider } from './utils/nonce-provider.ts'
 
 const ABORT_DELAY = 5000
 
 init()
 global.ENV = getEnv()
 
-export default async function handleRequest(
-	request: Request,
-	responseStatusCode: number,
-	responseHeaders: Headers,
-	remixContext: EntryContext,
-) {
-	const { currentInstance, primaryInstance } = await getInstanceInfo()
+type DocRequestArgs = Parameters<HandleDocumentRequestFunction>
 
+export default async function handleRequest(
+	...args: DocRequestArgs
+) {
+  const [
+    request,
+    responseStatusCode,
+    responseHeaders,
+    remixContext,
+    loadContext,
+  ] = args
+	const { currentInstance, primaryInstance } = await getInstanceInfo()
 	responseHeaders.set('fly-region', process.env.FLY_REGION ?? 'unknown')
 	responseHeaders.set('fly-app', process.env.FLY_APP_NAME ?? 'unknown')
 	responseHeaders.set('fly-primary-instance', primaryInstance)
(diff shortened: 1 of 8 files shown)


server/index.ts
```diff
@@ -26,32 +29,76 @@ app.disable('x-powered-by')
...
+        'font-src': ["'self'"],
+        'frame-src': [
+          "'self'",
+        ],
+        'img-src': [
+          "'self'",
+          'data:',
+          'user-images.githubusercontent.com',
+          'raw.githubusercontent.com',
+          ...(MODE === 'development' ? ['cloudflare-ipfs.com'] : []),
```
kentcdodds, line 63: I think I'd like to move the images to be served by the app server instead so we don't have to include this in the csp.

server/index.ts
```diff
@@ -26,32 +29,76 @@ app.disable('x-powered-by')
...
+        ],
+        'media-src': ["'self'", 'res.cloudinary.com', 'data:', 'blob:'],
```
kentcdodds, line 65: Same for this, let's limit this as much as possible.

app/entry.server.tsx
```diff
@@ -29,24 +35,33 @@ export default async function handleRequest(
...
+          const body = new PassThrough()
 					responseHeaders.set('Content-Type', 'text/html')
+          const dataEvtTransform = new Transform({
+            transform(chunk, _, callback) {
+              const string = chunk.toString()
+              const replaced = string.replace(/data-evt-/g, `nonce="${nonce}" `)
+              callback(null, replaced)
+            },
+          })
+          pipe(dataEvtTransform).pipe(body)
```
kentcdodds, line 58: Let's not bother with this. This is something I'm doing in my own site that's unnecessary for anyone else.

app/root.tsx
```diff
@@ -125,14 +127,16 @@ export default function App() {
 					<ThemeSwitch />
 				</div>
 				<div className="h-5" />
-				<ScrollRestoration />
-				<Scripts />
+				<ScrollRestoration nonce={nonce}/>
+				<Scripts nonce={nonce}/>
 				<script
+          nonce={nonce}
+          suppressHydrationWarning
```
kentcdodds, line 134: Is this necessary? It should not be.

kentcdodds (changes requested):
Thanks for working on this!

app/root.tsx
```diff
@@ -94,7 +96,7 @@ export default function App() {
 				<meta name="viewport" content="width=device-width,initial-scale=1" />
 				<Links />
 			</head>
-			<body className="flex h-full flex-col justify-between bg-night-700 text-white">
+			<body suppressHydrationWarning className="flex h-full flex-col justify-between bg-night-700 text-white">
```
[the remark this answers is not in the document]
paolapog: My fault, I was suppressing an error from a Chrome's extensions (I didn't find a clue yesterday), I'll remove it 💪🏼

server/index.ts
```diff
@@ -26,32 +29,76 @@ app.disable('x-powered-by')
...
 )
-
 // Everything else (like favicon.ico) is cached for an hour. You may want to be
 // more aggressive with this caching.
 app.use(express.static('public', { maxAge: '1h' }))
 
-morgan.token('url', (req, res) => decodeURIComponent(req.url ?? ''))
 app.use(morgan('tiny'))
 
+morgan.token('url', (req, res) => decodeURIComponent(req.url ?? ''))
```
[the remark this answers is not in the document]
paolapog: I made a `git rebase -i main` and I took the last changes: it's already present in the app

app/root.tsx
kentcdodds: If you're getting that warning then something's wrong. This should all work without any warnings.

paolapog: I was looking on the other project (kentcdodds.com) and this attribute is present. Maybe there is the same problem in the current implementation
I'll try to make it work 💪🏼

paolapog: update: https://react.dev/reference/react-dom/client/hydrateRoot 
`In development mode, React warns about mismatches during hydration.`
I tried with `NODE_ENV=production` and it doesn't complain 👍🏼 so I think I can remove it without concerns

kentcdodds (approved):
Looks great. Thanks a lot!

Issue #68, opened 2023-05-19
Fly volume in wrong region


joshmedeski:
I've been following the README and when I deployed I got this error.

> Error: error creating a new machine: failed to launch VM: Mounts source volume "vol_lgz1vppx35yv78m3" is in the wrong region ("dfw" != "iad")

So, I created a new volume in `iad` and I got this message:

> Error: error creating a new machine: failed to launch VM: Mounts source volume "vol_1q85vg7xkgzvzdxe" is in the wrong region ("iad" != "dfw")

What's happening? When I created the apps I wasn't asked for a region and when I created the volumes I chose `dfw` for both.

joshmedeski:
I just deleted my apps and volumes and tried again with a new name and I got this:

> Error: error creating a new machine: failed to launch VM: Mounts source volume "vol_915grnw3g9krn70q" is in the wrong region ("dfw" != "sjc")

Not sure what's going on. I'd love to get the Epic Stack up and running before committing to the stack to develop my next product.

kentcdodds:
Hi @joshmedeski,

I'm afraid there's not much I can do to help you in this. I would suggest you go to https://community.fly.io to see what's up with your app. They're pretty responsive.

chessydk:
I had the same problem and found out that you need to specify your primary region in the fly.toml file. https://fly.io/docs/reference/configuration/#primary-region

The error comes because connected volumes and machines have to be in the same region!

PR #84, opened 2023-05-22
Set default region for both machines and volumes

chessydk:
It seems like for some fly users (me and also reported by @joshmedeski here: https://github.com/epicweb-dev/epic-stack/issues/68), that the default region for machines and volumes are not the same and a deploy will fail because machine and mounted volume needs to be in same region. Specifying them explicitly solves the problem. I picked "sjc" a bit randomly, it could just as well be another, but it will never fit all users anyway.

Documentation for "primary_region": https://fly.io/docs/reference/configuration/#primary-region

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Prior to your first deployment, you'll need to do a few things:
   make sure you change the `primary_region` in fly.toml as well):
 
   ```sh
-  fly volumes create data --region sjc --size 1  --app epic-stack-template
+  fly volumes create data --region sjc --size 1 --app epic-stack-template
   fly volumes create data --region sjc --size 1 --app epic-stack-template-staging
   ```
 
--- a/README.md
+++ b/README.md
@@ -199,11 +199,13 @@ Prior to your first deployment, you'll need to do a few things:
 
 - Create a persistent volume for the sqlite database for both your staging and
   production environments. Run the following (feel free to change the GB size
-  based on your needs):
+  based on your needs and the region of your choice
+  (`https://fly.io/docs/reference/regions/`). NB. If you do change the region,
+  make sure you change the `primary_region` in fly.toml as well):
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -199,11 +199,13 @@ Prior to your first deployment, you'll need to do a few things:
...
   production environments. Run the following (feel free to change the GB size
-  based on your needs):
+  based on your needs and the region of your choice
+  (`https://fly.io/docs/reference/regions/`). NB. If you do change the region,
+  make sure you change the `primary_region` in fly.toml as well):
 
   ```sh
-  fly volumes create data --size 1 --app epic-stack-template
-  fly volumes create data --size 1 --app epic-stack-template-staging
+  fly volumes create data --region sjc --size 1  --app epic-stack-template
```
kentcdodds, line 207:
```suggestion
  fly volumes create data --region sjc --size 1 --app epic-stack-template
```

README.md
```diff
@@ -199,11 +199,13 @@ Prior to your first deployment, you'll need to do a few things:
 
 - Create a persistent volume for the sqlite database for both your staging and
   production environments. Run the following (feel free to change the GB size
-  based on your needs):
+  based on your needs and the region of your choice
+  (`https://fly.io/docs/reference/regions/`). NB. If you do change the region,
```
kentcdodds, line 203: I don't know what "NB" means here, but I think it makes sense without it.

```suggestion
  (`https://fly.io/docs/reference/regions/`). If you do change the region,
```

kentcdodds (approved):
I think this is a good temporary solution. Hopefully later we can remove this.

README.md
chessydk: Yep, works fine without. N.B. was "Abbreviation of Latin nota bene. Used before a piece of important information to make readers notice it."

PR #92, opened 2023-05-24
reload server build only after it was changed

onemen:
Currently the code clear cache and import the build on each request for development and production

We only need to clear cache and import the build on development when the build file change.

You can also look at [remix express template](https://github.com/remix-run/remix/blame/templates_v2_dev/templates/express/server.js)


kentcdodds (approved):
Thanks!

Issue #102, opened 2023-05-26
Zod: async function during sync operation


andrecasal:
The signup page is showing this error:

<img width="1029" alt="Screenshot 2023-05-26 at 11 01 31" src="https://github.com/epicweb-dev/epic-stack/assets/22457463/e99d37e2-a7ba-4602-b2b9-096edf8a4714">

I'm not familiar enough with Conform and Zod to know how to fix this, but I think this pertains to the `isEmailUnique()` function being async, but the client-side validation running sync.

edmundhung:
There is a bug on the [signup schema](https://github.com/epicweb-dev/epic-stack/blob/main/app/routes/_auth%2B/signup.tsx#L24-L30) that set a default implementation for `isEmailUnique`.

It should be optional instead:
```ts
function createSchema(
  constraints: {
    isEmailUnique?: (email: string) => Promise<boolean>
  } = {},
)
```

andrecasal:
I think we also need to add a `return` here:
```tsx
if (typeof constraints.isEmailUnique === 'undefined') {
	ctx.addIssue({
		code: z.ZodIssueCode.custom,
		message: conform.VALIDATION_UNDEFINED,
	})
	return // 👈 add return statement
}
```

PR #105, opened 2023-05-26
Refine conform setup

edmundhung:
This fixes the async validation setup on the signup form as mentioned on #102 with an improvement on how submission intent is handled:

As Conform is using the action for async validation as well, the status on the action data was interpreted as an error and triggered a red cross to be displayed on the submit button before the user actually submit the form. It's now returning the submission with an idle status instead until an actual submission is made.

## Test Plan

- Go to `/signup`
- Fill in `kody@kcd.dev` on the email
- An error should be shown once the field is unfocused
- Typing another email
- The error should be gone once the field is unfocused again

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
This makes sense to me 👍 Thanks!

PR #106, opened 2023-05-27
feat: progressively enhanced checkbox field

edmundhung:
## Problem

If we submit the onboarding form without agreeing the terms of service and privacy policy, an error message will be shown. However, this message will not be disappeared when we click on the checkbox and unfocus it even though the `shouldRevalidate` config is set to `onBlur`.

## Why?

Conform listens to native form events like the `input` or `blur` event to trigger validation. However, the checkbox component from Radix is implemented with a button element and an hidden checkbox. When we click on the button, a `click` event will be dispatched from the button element instead of an input event dispatching from the checkbox element. When we unfocus the button, there will also be a `blur` event dispatched from the button element instead of the checkbox element. 

Basically, Conform does not know when the user unfocus the checkbox.

## Solution

In this PR, we used the [useInputEvent](https://conform.guide/api/react#useinputevent) hook to emulate several native form events. You can also check the [guide](https://conform.guide/integrations) and some [examples](https://conform.guide/examples#ui-components) integrating different UI libraries. 

## Notes

I do not consider this coupling the input component with Conform. Like many custom input  renders an extra hidden input for better HTML Form support, we are just raising the bar on making custom input more aligns with native form control. That's why I call this a progressively enhanced checkbox 😄 .

Also, now we can listen to form event on the form without worrying whether it has any custom inputs as they both behave the same way!
  
```tsx
function example() {
  // Let's say we want to submit the form on change
  const submit = useSubmit();

  return (
    <form onChange={event => submit(event.target.form)}>
      {/* ... */}
    </form>
  );
}
```


app/utils/forms.tsx
```diff
@@ -98,25 +99,42 @@ export function CheckboxField({
...
 		<div className={styles.checkboxField}>
+			<input
+				{...buttonProps}
+				ref={ref}
+				id={undefined /* buttonProps.id should be assigned to the Radix checkbox instead */}
+				autoComplete="off"
+				className="sr-only"
+				aria-hidden
+				tabIndex={-1}
+				onFocus={() => buttonRef.current?.focus()}
```
edmundhung, line 122: Conform will focus on the first invalid input on submit. This allow us to forward it to the button

app/utils/forms.tsx
```diff
@@ -98,25 +99,42 @@ export function CheckboxField({
...
+			/>
```
edmundhung, line 123: We have to setup an extra input as there is no way to [access the checkbox element from the Radix Checkbox](https://github.com/radix-ui/primitives/discussions/874).

app/utils/forms.tsx
```diff
@@ -98,25 +99,42 @@ export function CheckboxField({
...
 	const fallbackId = useId()
 	const id = buttonProps.id ?? buttonProps.name ?? fallbackId
 	const errorId = errors?.length ? `${id}-error` : undefined
+	// To emulate naitve input events that Conform listen to:
+	// See https://conform.guide/integrations
+	const [ref, control] = useInputEvent()
+	const buttonRef = useRef<HTMLButtonElement>(null)
 	return (
 		<div className={styles.checkboxField}>
+			<input
```
kentcdodds, line 114: I'm concerned about this. Won't this mean we're rendering two <input /> elements per checkbox?

kentcdodds (commented):
Thanks a lot for the PR! Just one question.

app/utils/forms.tsx
edmundhung: Yes. But only this checkbox element will have the name attribute set. The one rendered by Radix will have no name and will be ignored for FormData.

kentcdodds: Should we make this a part of a generic checkbox component so people don't have to do this song and dance every time they need a checkbox?

edmundhung: Sure. We are now modifying the `<CheckboxField />` component so people can reuse it. Or do you mean something else?

kentcdodds: Yep, that's what I'm thinking 👍

edmundhung:
Just resolved the conflict. Is there anything you wanna change in this PR?

app/utils/forms.tsx
kentcdodds: Could you add this as a comment in the code so people understand why we're doing what looks odd at first?

kentcdodds (changes requested):
Thank you for the help here. Sorry for the confusion on my part 😅 Just one more thing.

edmundhung:
I just figured out a better approach that could make this work without an extra checkbox element. I will create another PR later this week. 😄

kentcdodds:
Oh great! Thanks :)

PR #107, opened 2023-05-27
Use LiteFS supervisor, replace start script

mattfysh:
I believe the existing start.js script can be replaced by the LiteFS supervisor feature.

I also have an inkling there may be a few latent issues when you try to go multi-node and/or multi-region. In the more complex topologies, can a node that was just deployed be the primary at startup, ie. once `litefs mount` is finished setting up?

From what I gather, without the use of the `promote` flag, the new node is not immediately elected primary so the migration script may never run

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds:
Thanks a lot @mattfysh! This would definitely be an improvement!

@benbjohnson, could you help with this one? I want to make sure we're doing things right here since I've not heard of the `promote` or `if-primary` options.

benbjohnson:
Yes, those `litefs.yml` changes look correct.

- If `promote` is `true`, it will automatically promote the node to primary after it syncs with the cluster. 

- The `if-candidate` flag on the `exec` subcommands will only run the command on candidate nodes when it starts up. This, combined with `promote`, means you can have commands like migrations automatically run on the primary node when you deploy.

> can a node that was just deployed be the primary at startup, ie. once litefs mount is finished setting up?

Yes, a new node can become primary but the promotion only occurs after it has sync'd with the cluster so it will have the full database.

other/litefs.yml
```diff
@@ -21,12 +21,16 @@ proxy:
 # You must specify "experiement.enable_consul" for FLY_CONSUL_URL to be available.
 lease:
   type: 'consul'
-  candidate: true
-  # If you'd like to control the candidate's possible region, you can use some
-  # simple expressions for this value, like so:
-  # candidate: ${FLY_REGION == 'sjc'}
+  candidate: ${FLY_REGION == 'sjc'}
```
benbjohnson, line 24: For new apps on Fly.io, we have a `PRIMARY_REGION` environment variable available so you can do:

```
candidate: ${FLY_REGION == PRIMARY_REGION}
```

mattfysh:
Thanks @benbjohnson! I've made the update to litefs.yml based on your feedback

For my own learning, if this repo were launched in its current state (before this PR) would the migration script ever be run? There is only a very short period during node start where it checks to see if it is the primary and performs the migration. If there are other candidates in the cluster, and the config is not using auto-promotion, wouldn't you miss the opportunity to run the migration script due to new nodes never being elected leader on startup?

benbjohnson:
You're correct. There's definitely a race in the previous approach so that the migration wouldn't always get run in a multi-candidate cluster.

Separately, migrations are tricky in distributed systems—even in the current approach. Even when the migration is applied to the primary, that schema change will propagate to other replica nodes before they are deployed to the new version of the app (in a rolling deploy situation). It's common to see people apply migration scripts out-of-band of their deployments so they're not so tied to a particular version. Kent's used that approach in the past as well.

mattfysh:
You're right, I'm often left scratching my head when it comes to versioning + deployments. I think the first step would be to have the migration script running predictably, and then I'd approach this with a widen/narrow multi-deployment where your goal is to get your database from A -> B

1. Widen app to consume A or B
2. Widen db to provide A and B
3. Narrow app to consume B
4. narrow db to provide B

There is a higher cost to this approach due to having to perform multiple deployments and wait for a settled state between each. However in some cases, some of these steps can be combined

Either way, I don't think it's something you can automate away just yet ... still waiting on that migration nirvana 🙏

benbjohnson:
Yeah, a widen/narrow approach is a good way to view it. In practice, I find that most schema changes work across application versions (e.g. add a column) so it's not too onerous. But in more complex cases, it can definitely be a head scratcher! :)

kentcdodds:
I'll get this merged soon.

We'll want documentation about the widen narrow approach as well.

kentcdodds:
Thanks a lot both of you!

kentcdodds:
The deploy of this change failed unexpectedly 🤔

https://github.com/epicweb-dev/epic-stack/actions/runs/5126148176/jobs/9220302403#step:6:239

image

kentcdodds:
I guess litefs doesn't like `&&` in the `cmd` option 😅 I've just [removed the echo](https://github.com/epicweb-dev/epic-stack/commit/71b948c7395972d34e331b578f7abae449691f1c). It's fine.

mattfysh:
Nice catch, that'll teach me not to sneak in a "small" change after PR ;)

benbjohnson:
> litefs doesn't like `&&`

Yeah, the `&&` is a `bash` thing. You'd have to call it through bash. Maybe something like:

```
cmd: "bash -c 'echo "Starting app..." && exec npm start'"
```

That _should_ run the `echo` and then `exec` replaces the `bash` process with `npm start`.

You could also just add two `cmd` items too. :)

PR #109, opened 2023-05-28
improve form types

lifeiscontent:
## Summary

in a recent patch release of conform-to they've updated the types to be able to include the transformations of a form input and output, this improves the type safety of the form submissions.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
     "/server-build"
   ],
   "dependencies": {
-    "@conform-to/react": "^0.6.1",
-    "@conform-to/zod": "^0.6.1",
+    "@conform-to/react": "0.6.3",
+    "@conform-to/zod": "0.6.3",
     "@prisma/client": "^4.14.1",
     "@radix-ui/react-checkbox": "^1.0.4",
     "@radix-ui/react-dialog": "^1.0.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -35,8 +35,8 @@
     "/server-build"
   ],
   "dependencies": {
-    "@conform-to/react": "^0.6.1",
-    "@conform-to/zod": "^0.6.1",
+    "@conform-to/react": "0.6.3",
```
lifeiscontent, line 38: @kentcdodds I'm not sure if there's a reason for not using pinned versions as its would provide stronger gaurentees that the application still works as expected, e.g. in this patch release the types actually changed so if we were to run an upgrade even though it wasn't specified, it would of resulted in a breaking change for this application.

edmundhung:
The `useForm` hook should manage to inferred it from both `onValidate` (output type) and `constraints` (input type) as well. I think the question here is if we  want to type it explicitly.

lifeiscontent:
@edmundhung I didn't realize it was inferred, I've updated the code to use the inferred types. Thanks for the context! :)

kentcdodds:
This PR now appears to not do anything about the form types. We'll be upgrading to the latest version of conform soon to address some ESM issues so I'm going to close this one. Thanks for the help anyway!

PR #111, opened 2023-05-30
simplify createRequestHandler

onemen:
1. `getRequestHandlerOptions` is not a Promise in our case.

2. because `devBuild` and `build` are both of type `ServerBuild` we don't need to wait for createRequestHandler.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -8,7 +8,7 @@ import address from 'address'
 import closeWithGrace from 'close-with-grace'
 import helmet from 'helmet'
 import crypto from 'crypto'
-import { createRequestHandler } from '@remix-run/express'
+import { type RequestHandler, createRequestHandler } from '@remix-run/express'
 import { type ServerBuild, broadcastDevReady } from '@remix-run/node'
 import getPort, { portNumbers } from 'get-port'
 import chalk from 'chalk'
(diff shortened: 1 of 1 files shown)


server/index.ts
```diff
@@ -72,27 +72,15 @@ app.use(
...
 	}
-	return { build, mode: MODE, getLoadContext }
+	return createRequestHandler({ build, mode: MODE, getLoadContext })
 }
 
-app.all(
-	'*',
-	process.env.NODE_ENV === 'development'
-		? async (req, res, next) => {
-				return createRequestHandler(await getRequestHandlerOptions(devBuild))(
```
kentcdodds, line 88: This change won't work because we need to make certain every request uses the latest version of the `devBuild` variable. With the changes here it will only ever use the first version.

kentcdodds (changes requested):
Thanks! I wish it could be this easy!

onemen:
how about this
```typescript
function getRequestHandler(build: ServerBuild): RequestHandler {
	function getLoadContext(_: any, res: any) {
		return { cspNonce: res.locals.cspNonce }
	}
	return createRequestHandler({ build, mode: MODE, getLoadContext })
}

app.all(
	'*',
	process.env.NODE_ENV === 'development'
		? async (req, res, next) => {
				return getRequestHandler(devBuild)(req, res, next)
		  }
		: getRequestHandler(build),
)
```

kentcdodds:
Looks good to me 👍

PR #113, opened 2023-05-31
simplify createRequestHandler

onemen:
2nd try after #111

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -8,7 +8,7 @@ import address from 'address'
 import closeWithGrace from 'close-with-grace'
 import helmet from 'helmet'
 import crypto from 'crypto'
-import { createRequestHandler } from '@remix-run/express'
+import { type RequestHandler, createRequestHandler } from '@remix-run/express'
 import { type ServerBuild, broadcastDevReady } from '@remix-run/node'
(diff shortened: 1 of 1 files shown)


imcodingideas:
@onemen can you update your PR with a test plan to validate this works?

onemen:
```
const Test = x => 42
```

kentcdodds (approved):
Looks great to me 👍

kentcdodds:
> ```
> const Test = x => 42
> ```

I think @imcodingideas was trying to be genuinely helpful, so a short joke response like this is not appropriate.

@imcodingideas, @onemen has contributed to this stack quite a bit already so we've got an established relationship and also on this specific issue we've discussed it already.

That said, I do think it's a good idea to get in the habit of demonstrating things still work properly. Luckily we've got tests in place and for something like this, all the playwright tests would fail if it wasn't working so I think we're ok.

Although, now that I think about it, this could potentially impact development (HMR/HDR) which we don't have any automated tests for (and really probably couldn't reasonably do), so having a screenshot or video demonstrating that's still working would be appropriate.

All that said, I'm confident that this will work fine so I'm going to merge. Thanks friends!

onemen:
got it 👍

PR #114, opened 2023-05-31
Create redirects.md

jacobparis:
Documenting how redirects are handled in the Epic Stack, including code examples for HTTPS, removing trailing slashes, and forcing www subdomains

## Checklist

- [x] Docs updated

### DIFF
--- a/docs/redirects.md
+++ b/docs/redirects.md
@@ -44,7 +44,9 @@ You can redirect root domain traffic to your www subdomain or vice versa.
 
 [DNS level redirects do not work with Fly](https://community.fly.io/t/how-to-redirect-from-non-www-to-www/5795/2). The recommended way to implement this redirect is in your application code.
 
-First, you will need to register SSL certificates for both the www and the root domain. That allows traffic from either source to hit your application, and then you can redirect to your preferred format in code. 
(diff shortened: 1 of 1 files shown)


docs/redirects.md
```diff
@@ -0,0 +1,71 @@
...
+})
+```
+
+## www subdomains
+
+You can redirect root domain traffic to your www subdomain or vice versa.
+
+[DNS level redirects do not work with Fly](https://community.fly.io/t/how-to-redirect-from-non-www-to-www/5795/2). The recommended way to implement this redirect is in your application code.
+
+First, you will need to register SSL certificates for both the www and the root domain. That allows traffic from either source to hit your application, and then you can redirect to your preferred format in code.
```
kentcdodds, line 47: Could you link to some docs explaining how to do this on fly?

kentcdodds (commented):
This is awesome! Just one thing.

docs/redirects.md
jacobparis: Updated

Fly has docs for doing this with the CLI but unfortunately none for using the dashboard (which is the way I've always done it)

I don't want to go into too much detail about their dashboard as that becomes a maintenance burden (perhaps the same reason they chose not to do that too) but this should be enough info to get the user in the right direction

MichaelDeBoey (approved):

kentcdodds (approved):
Thanks a million!

PR #115, opened 2023-05-31
refine username validation to not be able to use an email as username

Mat-moran:
An user can use an email as usernam. This can conflict when using method that takes an argument as username or password and find the user for any of them. 

## Test Plan

I have verify locally that the regex is OK using an username with `@` gives a zod error and if the username does not contain `@` it works as expected.

### DIFF
--- a/app/utils/user-validation.ts
+++ b/app/utils/user-validation.ts
@@ -4,7 +4,7 @@ export const usernameSchema = z
 	.string()
 	.min(3, { message: 'Username is too short' })
 	.max(20, { message: 'Username is too long' })
-	.regex(/^[^@]*$/, "username can not contains @")
+	.regex(/^[^@]*$/, 'Username can not contain @')
 export const passwordSchema = z
 	.string()
 	.min(6, { message: 'Password is too short' })
--- a/app/utils/user-validation.ts
+++ b/app/utils/user-validation.ts
@@ -4,6 +4,7 @@ export const usernameSchema = z
 	.string()
 	.min(3, { message: 'Username is too short' })
 	.max(20, { message: 'Username is too long' })
+	.regex(new RegExp("^[^@]*$"), "username can not contains @")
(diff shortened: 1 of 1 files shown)


app/utils/user-validation.ts
```diff
@@ -4,6 +4,7 @@ export const usernameSchema = z
 	.string()
 	.min(3, { message: 'Username is too short' })
 	.max(20, { message: 'Username is too long' })
+	.regex(new RegExp("^[^@]*$"), "username can not contains @")
```
kentcdodds, line 7: Wouldn't this work?

```suggestion
	.regex(/@/, 'Username can not contain @')
```

Mat-moran: does not work, we can avoid using `new RegExp(regular expression)` but only `/@/` is telling zod that the username should have the symbol `@`

app/utils/user-validation.ts
```diff
@@ -4,6 +4,7 @@ export const usernameSchema = z
 	.string()
 	.min(3, { message: 'Username is too short' })
 	.max(20, { message: 'Username is too long' })
+	.regex(/^[^@]*$/, "username can not contains @")
```
kentcdodds, line 7:
```suggestion
	.regex(/^[^@]*$/, 'Username can not contain @')
```

app/utils/user-validation.ts
```diff
@@ -4,6 +4,7 @@ export const usernameSchema = z
 	.string()
 	.min(3, { message: 'Username is too short' })
 	.max(20, { message: 'Username is too long' })
+	.regex(/^[^@]*$/, 'Username can not contain @')
```
kentcdodds, line 7:
```suggestion
	.regex(/^[^@]*$/, 'Username can not contain "@"')
```

kentcdodds (approved):
Thanks!

PR #116, opened 2023-06-01
update all the things

onemen:
## dependencies/devdependencies
 * update all
 * remove unused
 * add missing

## Test Plan
all test are passing 

## Checklist

- [x] Tests updated
- [ ] Docs updated


app/routes/_auth+/signup.tsx
```diff
@@ -156,10 +160,10 @@ export default function SignupRoute() {
 	const signupFetcher = useFetcher<typeof action>()
 	const [form, fields] = useForm({
 		id: 'signup-form',
-		constraint: getFieldsetConstraint(createSchema()),
+		constraint: getFieldsetConstraint(emailFieldSchema),
 		lastSubmission: signupFetcher.data?.submission,
 		onValidate({ formData }) {
-			return parse(formData, { schema: createSchema() })
+			return parse(formData, { schema: emailFieldSchema })
```
onemen, line 166: @kentcdodds 

without this change `onboarding.test.ts` was failed, I did not find in `conform-to/react` a way to use asunc `createSchema` here.

the test work and i've add a test for `user already exists` case

tests/memoize-unique.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * based on https://github.com/faker-js/faker/issues/1785#issuecomment-1407773744
+ * replace this code with memoize-unique when it is published
+ */
```
onemen, line 4: @kentcdodds 

faker.helpers.unique will be remove in the `faker` next version.
this is the replacement proposed by one of `faker` maintainer

package.json
```diff
@@ -35,100 +35,95 @@
...
     "tailwindcss-radix": "^2.8.0",
     "tsconfig-paths": "^4.2.0",
     "tsx": "^3.12.7",
     "typescript": "^5.0.4",
-    "vite": "^4.3.5",
+    "vite": "^4.3.9",
     "vite-tsconfig-paths": "^4.2.0",
-    "vitest": "^0.31.0",
+    "vitest": "^0.31.4",
     "vitest-mock-extended": "^1.1.3"
```
onemen, line 127: @kentcdodds,

this is `npx depcheck` output, maybe we can remove some ?
```
Unused dependencies
* remix-auth-socials
Unused devDependencies
* @testing-library/dom
* @testing-library/react
* @types/eslint
* @vitejs/plugin-react
* @vitest/coverage-c8
* autoprefixer
* c8
* postcss
* prettier-plugin-sql
* prettier-plugin-tailwindcss
* tailwindcss
* tailwindcss-radix
* tsconfig-paths
* vitest-mock-extended
```

tests/e2e/onboarding.test.ts
```diff
@@ -38,9 +38,18 @@ test('onboarding', async ({ page }) => {
...
 	const emailTextbox = page.getByRole('textbox', { name: /email/i })
+
+	// Check for the error message
 	await emailTextbox.click()
-	await emailTextbox.fill(onboardingData.email)
+	await emailTextbox.fill('kody@kcd.dev')
+	await page.getByRole('button', { name: /launch/i }).click()
+	await expect(
+		page.getByText(/user already exists with this email/i),
+	).toBeVisible()
```
onemen, line 48: @kentcdodds,

this test work locally but failed on CI
let me know how to fix it

package.json
kentcdodds: This is tough. Some of these we don't need today and I'm fine removing them. Others are used, but depcheck can't tell (like prettier/tailwind/c8/vitest coverage stuff). While others aren't used in the project but people will need when they start building out features (like testing library stuff).

I think the only thing we can safely remove from that list is remix-auth-socials.

tests/e2e/onboarding.test.ts
kentcdodds: Why was this change made?

onemen: I've just added it to make sure it wok, I can remove it if you don't want it

kentcdodds: Yes, it's possible: https://github.com/nektos/act

I don't think we want to write the test this way. It has an implicit dependency on the seed script. If we wanted to test this we'd want to make a new user first and then use that one. I don't think that's necessary for this level of testing though. Most of the time E2E should test happy-path stuff. If we want to test this we should go with lower level tests (like unit or integration tests).

onemen: ok, I will remove it

kentcdodds (approved):
Thank you!

PR #118, opened 2023-06-01
Add tailwind merge

rajeshg:
Replace clsx with tailwind-merge
See https://github.com/epicweb-dev/epic-stack/discussions/117
## Test Plan

Current tests should pass
## Checklist

- [x] Tests updated - Not applicable. Current tests should suffice
- [x] Docs updated - utility function. 

## Screenshots

Not applicable.


app/utils/misc.ts
```diff
@@ -1,4 +1,6 @@
+import { ClassNameValue } from 'node_modules/tailwind-merge/dist/lib/tw-join.js'
```
sergiodxa, line 1: I don’t think relying on internals of a lib is a good idea

app/utils/misc.ts
```diff
@@ -1,4 +1,6 @@
...
 import React from 'react'
+import { twMerge, twJoin } from 'tailwind-merge'
```
sergiodxa, line 3: I think instead of twJoin it should be using clsx, the docs of tailwind-merge says that twJoin is similar expect it doesn’t support object syntax that clsx supports, and if it already use clsx I see no reason to remove a possible feature.

kentcdodds: I have mixed feelings about this. Their reasoning for no object syntax is pretty good, but I really hate abusing `&&` like that. I'm going to think about this more. I don't like having two ways to do the same thing 😕

sergiodxa: More than two ways people should just use consolidate, maybe it be a separate package that combines both if you don’t want to have clsx and tailwind-merge installed together.

app/utils/misc.ts
```diff
@@ -101,3 +103,7 @@ export function useDoubleCheck() {
 
 	return { doubleCheck, getButtonProps }
 }
+
+export function consolidate(...inputs: ClassNameValue[]) {
+	return twMerge(twJoin(inputs))
```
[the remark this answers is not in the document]
rajeshg: The idea could be to do a simple string join first using `twJoin`, followed by merging conflicts between tailwind styles using `twMerge`. May be twJoin has it's benefits when there are 100s of classes involved. But for the typical use cases, twMerge should be good enough. Can I go ahead and get rid of `twJoin` then?

app/utils/misc.ts
rajeshg: Add a local ClassNameValue type instead of relying on the internals

app/utils/misc.ts
```diff
@@ -101,3 +102,8 @@ export function useDoubleCheck() {
 
 	return { doubleCheck, getButtonProps }
 }
+
+type ClassNameValue = string[] | string | null | undefined | 0 | false;
+export function consolidate(...inputs: ClassNameValue[]) {
+	return twMerge(inputs)
```
kentcdodds, line 108: If this is all this does, then I think it makes more sense to just use twMerge directly instead of having our own wrapper function.

kentcdodds (changes requested):

app/utils/forms.tsx
```diff
@@ -161,15 +161,16 @@ export function getButtonClassName({
...
-		[pillClassName]: size === 'pill',
-	})
+	const className = twMerge(
+		baseClassName,
+		variant === 'primary' && primaryClassName,
+		variant === 'secondary' && secondaryClassName,
+		size === 'xs' && extraSmallClassName,
+		size === 'sm' && smallClassName,
+		size === 'md' && mediumClassName,
+		size === 'md-wide' && [mediumWideClassName],
```
kentcdodds, line 171:
```suggestion
		size === 'md-wide' && mediumWideClassName,
```

kentcdodds (approved):
Thank you!

app/utils/forms.tsx
```diff
@@ -191,7 +192,7 @@ export function Button({
 	return (
 		<button
 			{...props}
-			className={clsx(
+			className={twMerge(
 				props.className,
 				getButtonClassName({ size, variant }),
 				'flex justify-center gap-4',
```
mohammedzamakhan, line 198: props.className should be last @kentcdodds

kentcdodds: You're correct. Mind making a pull request

onemen:
maybe we can use helper function to convert

`Object.entries(classList).filter(([key,val])=>val).map(([key])=>key)`

```ts
const classList = {
		[primaryClassName]: variant === 'primary',
		[secondaryClassName]: variant === 'secondary',
		[extraSmallClassName]: size === 'xs',
		[smallClassName]: size === 'sm',
		[mediumClassName]: size === 'md',
		[mediumWideClassName]: size === 'md-wide',
		[pillClassName]: size === 'pill',
	}
```
to

```
const baseClassList = [
	primaryClassName,
	secondaryClassName,
	extraSmallClassName,
	smallClassName,
	mediumClassName,
	mediumWideClassName,
	pillClassName,
]
```

now we can apply `twMerge` :
`twMerge(...baseClassList, extraClass)`

kentcdodds:
If we're going to do that we may as well just keep clsx. It's really not that big.

PR #123, opened 2023-06-02
notes e2e test

rajeshg:
Add e2e test for notes basic features
## Test Plan

Run playwright tests
## Checklist

- [x] Tests updated
- [x] Docs updated

### DIFF
--- a/tests/e2e/notes.test.ts
+++ b/tests/e2e/notes.test.ts
@@ -10,16 +10,16 @@ test('Users can create notes', async ({ login, page }) => {
 
 	// blank form submission should result in errors
 	await page.getByRole('button', { name: /submit/i }).click()
-	let countErrors = page
-		.getByText('String must contain at least 1 character(s)')
-		.count()
-	await expect(countErrors).toEqual(2)
+	// count errors
+	await expect(
+		page.getByText('String must contain at least 1 character(s)'),
+	).toHaveCount(2)
 
 	// fill in form and submit
 	await page.getByRole('textbox', { name: /title/i }).fill(newNote.title)
-	await page.getByLabel(/^content$/i).fill(newNote.content)
+	await page.getByRole('textbox', { name: /content/i }).fill(newNote.content)
 
-	await page.getByRole('button').last().click()
+	await page.getByRole('button', { name: /submit/i }).click()
 	await expect(page).toHaveURL(new RegExp(`/users/${user.username}/notes/.*`))
 })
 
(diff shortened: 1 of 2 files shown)


tests/e2e/notes.test.ts
```diff
@@ -0,0 +1,74 @@
+import { faker } from '@faker-js/faker'
+import { expect, test } from '../playwright-utils.ts'
+
+test('Users can create notes', async ({ login, page }) => {
+	const user = await login()
+	await page.goto(`/users/${user.username}/notes`)
+
+	const newNote = createNewNote()
+	await page.locator("//a[text()='+ New Note']").click()
```
kentcdodds, line 9: Let's swap from `.locator` to the testing-library inspired API playwright added a while ago. Check the recommended locators here: https://playwright.dev/docs/locators#quick-guide

tests/e2e/notes.test.ts
```diff
@@ -0,0 +1,74 @@
...
+	// blank form submission should result in errors
+	await page.locator('button[type="submit"]').click()
+	await expect(page.locator('ul#note-editor-title-error')).toBeVisible()
+	await expect(page.locator('ul#note-editor-content-error')).toBeVisible()
+
+	// fill in form and submit
+	await page.locator('input[name="title"]').fill(newNote.title)
+	await page.locator('textarea[name="content"]').fill(newNote.content)
+	await page.locator('button[type="submit"]').click()
+	await expect(page).toHaveURL(new RegExp(`/users/${user.username}/notes/*`))
```
kentcdodds, line 20: Should this be:

```suggestion
	await expect(page).toHaveURL(new RegExp(`/users/${user.username}/notes/.*`))
```

Same thing below

tests/e2e/notes.test.ts
```diff
@@ -0,0 +1,74 @@
...
+})
+
+test('Users can edit notes', async ({ login, page }) => {
+	const user = await login()
+	await page.goto(`/users/${user.username}/notes`)
```
kentcdodds, line 25: I noticed the two tabs here. I suggest you install a prettier plugin for your editor so you get format on save. It saves a lot of time!

kentcdodds (changes requested):
This is very cool! Thanks for working on this.

I'm thinking we may want to remove this when people generate the project because while it's nice for us to have to verify we don't break stuff, it's just another thing for people to delete when they start building their app.

tests/e2e/notes.test.ts
rajeshg: Thanks! Removed pretty much all `.locator` usage, except a couple of instances where I needed to verify if deletion works as intended.

kentcdodds:
It looks like there's a lot more unnecessary use of `.locator`.

kentcdodds (approved):
Looks great! Thank you 👏

PR #131, opened 2023-06-03
ui(root): always render white user name color

sean-beard:
Just started a new project and noticed there is a color contrast issue when rendering the user name in light mode.

### Before
Screen Shot 2023-06-03 at 5 02 18 PM

### After
Screen Shot 2023-06-03 at 5 01 43 PM


kentcdodds (approved):
Thanks! There are a lot of such issues we need to deal with 😅

PR #133, opened 2023-06-04
start tsx watch as child process

onemen:
calling `tsx watch` as a child process give us more control on its environment and stdio.

- this will solve issue [#6504](https://github.com/remix-run/remix/issues/6504)
- one less npm script

## Test Plan

all test are working as before

## Checklist

- [ ] Tests updated
- [ ] Docs updated


package.json
```diff
@@ -10,8 +10,7 @@
     "build": "run-s build:*",
     "build:remix": "remix build",
     "build:server": "tsx ./other/build-server.ts",
-    "dev": "remix dev -c \"npm run dev:server\" --no-restart",
-    "dev:server": "cross-env MOCKS=true tsx watch --clear-screen=false --ignore \"app/**\" --ignore \"build/**\" --ignore \"node_modules/**\" --inspect ./index.js",
+    "dev": "remix dev -c \"cross-env MOCKS=true node --inspect ./server/dev-server.js\" --no-restart",
```
kentcdodds, line 13: Let's move the environment variable to the execa can and the --inspect flag too. Then we can do:

```suggestion
    "dev": "remix dev -c \"node ./server/dev-server.js\" --no-restart",
```

onemen: done

PR #134, opened 2023-06-04
adjust text color for light mode

onemen:
for current design text color is most of the time  white

## Test Plan
no test

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots
notes-light-before
notes-light-after


kentcdodds:
We should probably change the design of this page to not have so much dark on the light background

onemen:
this is just first aid for now

PR #138, opened 2023-06-05
Preload CSS and aggressive font caching

xHomu:
A couple performance tweaks:

* By default, HTML rendering is paused until CSS stylesheets streamed in! We can fix this by preload CSS as a resource, 
* Currently, fonts are cached only a hour like everything else in the Public folder, so let's set the cache header for fonts folder to a year instead.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -36,6 +36,7 @@ import { makeTimings, time } from './utils/timing.server.ts'
 
 export const links: LinksFunction = () => {
 	return [
+		// Preload CSS as a resource to avoid render blocking
 		{ rel: 'preload', href: fontStylestylesheetUrl, as: 'style' },
 		{ rel: 'preload', href: tailwindStylesheetUrl, as: 'style' },
 		cssBundleHref ? { rel: 'preload', href: cssBundleHref, as: 'style' } : null,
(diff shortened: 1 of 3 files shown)


app/root.tsx
```diff
@@ -55,8 +55,12 @@ export const links: LinksFunction = () => {
 		},
 		{ rel: 'manifest', href: '/site.webmanifest' },
 		{ rel: 'icon', href: '/favicon.ico' },
+		// Preload CSS as a resource to avoid render blocking
+		{ rel: 'preload', href: fontStylestylesheetUrl, as: 'style' },
 		{ rel: 'stylesheet', href: fontStylestylesheetUrl },
+		{ rel: 'preload', href: tailwindStylesheetUrl, as: 'style' },
 		{ rel: 'stylesheet', href: tailwindStylesheetUrl },
+		cssBundleHref ? { rel: 'preload', href: cssBundleHref, as: 'style' } : null,
```
kentcdodds, line 63: Really doesn't feel like this should be necessary, or if it should, I would expect it to be better if we moved all preload tags to before the stylesheets come in?

Could you show some before/after metrics?

kentcdodds (commented):
Thanks! Just one thing.

xHomu:
> Really doesn't feel like this should be necessary, or if it should, I would expect it to be better if we moved all preload tags to before the stylesheets come in?
>
> Could you show some before/after metrics?

It's a bit verbose, but that line solves this particular warning in Lighthouse:

image

Details on preloading CSS: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload

The underlying theory is a bit over my head, but I do see it generally bump Lighthouse Performance by a point, and sometimes help with CLS.


Note: moving the preloads up to the top makes a lot of sense, so I pushed a

kentcdodds (approved):
Luckily I don't expect people to change these style imports much (ever) so I'm fine with it. Otherwise I think I may not be as jazzed about this 😅

Thanks!

PR #156, opened 2023-06-10
chore: remove `ENCRYPTION_SECRET` generation

MichaelDeBoey:
As mentioned in https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/014-totp.md#consequences, we don't need these anymore


kentcdodds (approved):
Thanks! Not sure how I missed this.

PR #161, opened 2023-06-11
fix: TOPT flow comments

scr2em:


app/utils/totp.server.ts
```diff
@@ -18,7 +18,7 @@
  * Here's the typical process:
  * 1. Generate a secret (crypto.randomBytes(32).toString('hex'))
  * 2. Generate the TOTP with that secret (generateTOTP(secret))
- * 3. Store the secret and the TOTP in the database along with the thing you're verifying (e.g. user email)
+ * 3. Store the secret, digits, period in the database along with the thing you're verifying (e.g. user email)
```
andrecasal, line 21: I think it should be "Store the secret, digits, **and** period in the database..." 👍

kentcdodds:
```suggestion
 * 3. Store the secret, digits, and period in the database along with the thing you're verifying (e.g. user email)
```

kentcdodds (approved):
Thanks!

PR #167, opened 2023-06-12
add prettier overrides for json files

onemen:
We discussed about it [here](https://github.com/epicweb-dev/epic-stack/discussions/159)


kentcdodds (approved):
Super 👍 Thank you!

PR #168, opened 2023-06-12
update dependencies and remove unused

onemen:
## removed
- remix-utils - leftover from 2fa draft
- @testing-library/dom - never used
- @testing-library/react - never used
- @vitest/coverage-c8 - replaced by @vitest/coverage-v8
- c8 - not in use
- vitest-mock-extended - not in use

## Test Plan

All test are passing...

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Question

@kentcdodds, currently `postcss` and `autoprefixer` are only used by `vitest`, remix handle postcss and tailwind for us.
we can remove both `postcss` and `autoprefixer` if we update `vitest.config.ts` like this:
```diff
	plugins: [react(), tsconfigPaths()],
+	css: {
+		postcss: {
+			plugins: [],
+		},
+	},
	test: {
```
what do you think?


kentcdodds:
Oh, and as far as the postcss stuff is concerned, we don't really get rid of the deps anywhere but in our package.json which I suppose is a win. Let's proceed with that suggestion 👍

onemen:
its ready

kentcdodds (approved):
Super. Thanks!

onemen:
@kentcdodds,

After reading Remix docs again regarding [postcss](https://remix.run/docs/en/1.17.1/guides/styling#postcss),
I think I have made a mistake by removing `autoprefixer`

kentcdodds:
Let's double-check that and add it back if it's needed

andrecasal:
There's no error if autoprefixer is removed, but is it required for prefixes to be inserted. Just ran into this now.

kentcdodds:
Well that's annoying! They should error out for that :( Fixing now.

PR #179, opened 2023-06-14
Fix fake user creation based on username rules

andrecasal:
Fix fake user creation based on username rules.

## Test Plan

All tests pass ✅


kentcdodds (approved):
Thanks!

PR #183, opened 2023-06-15
fix: report error

jansedlon:
## Test Plan
Hi! Many times I have experienced a situation where something happened during development and I couldn't see any error.
To give you an example.. I installed `lexical` and `@lexical/react`, which is a wysiwyg builder from Facebook. I think they don't provide ESM bundle or whatever. My point is that if you use it anywhere in the code like
```ts
import { LexicalComposer } from '@lexical/react/LexicalComposer.js';
console.log(LexicalComposer);
```

The build starts hanging
image

There is no error to be viewed. I dug deeper and had to debug the code. Turns out, the `closeWithGrace` is called in `tests/mocks/index.ts`, it stops the mock server but does nothing else.

Not sure if this PR is the right approach, but at least it will report any errors in the console and closes the whole dev server.

With this change, it looks like this
image

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -177,10 +177,22 @@ ${chalk.bold('Press Ctrl+C to stop')}
 	}
 })
 
-closeWithGrace(async () => {
+closeWithGrace(async ({ err }) => {
+	// log the error early
+	if (err) {
+		console.error(chalk.red(err))
+		console.error(chalk.red(err.stack))
+	}
+
+	// close up things
 	await new Promise((resolve, reject) => {
 		server.close(e => (e ? reject(e) : resolve('ok')))
 	})
+
+	// if there was an error, then exit with a failure code
+	if (err) {
+		process.exit(1)
+	}
 })
 
 // during dev, we'll keep the build module up to date with the changes
--- a/tests/mocks/index.ts
+++ b/tests/mocks/index.ts
@@ -1,3 +1,4 @@
+import chalk from 'chalk'
 import { rest } from 'msw'
 import { setupServer } from 'msw/node'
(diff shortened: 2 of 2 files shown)


kentcdodds (approved):
Good catch. I've moved and modified your changes to the server/index.ts which I think is the better place for this. Let me know if this doesn't solve your issue.

Thank you!

jansedlon:
I don't think that's gonna work, because when I debugged it, the server/index.ts wasn't run, that why I put it in the mock server file. I think I tried your approach at one point and didn't work, but I'm gonna have to double check tomorrow

kentcdodds:
Perhaps we need it in both 🤔 Or maybe we should put it in the `index.js` file in the root. Or all three... It just feels pretty strange putting it in the mocks directory.

onemen:
In my testing, on Windows 11, in case of an error in mock server or express server `closeWithGrace` log to the console and close the server however the tsx process does not close and have to manually close it.

andrecasal:
> In my testing, on Windows 11, in case of an error in mock server or express server `closeWithGrace` log to the console and close the server however the tsx process does not close. I have to manually close it.

I can confirm this also happens in the latest macOS (Ventura 13.4).

kentcdodds:
I've got work locally that puts it in the root index.js and that works for me.

andrecasal:
> I've got work locally that puts it in the root index.js and that works for me.

Sorry, I didn't get what you mean. Having it wrapping everything on the index.js enough?

kentcdodds:
This: https://github.com/epicweb-dev/epic-stack/commit/9543c5fb2b24f48f4f4b8313b8f2898b458ded99 :)

PR #185, opened 2023-06-15
Fix flaky logout

andrecasal:
`submit(e.currentTarget)` pushes the form's submission task to the macrotask queue in the event loop to process later. The click event bubbles up to `<DropdownMenu.Portal>` which will remove itself from the DOM by pushing the removal task to the macrotask queue.

The first task in the queue, the form's submission, is sent to the call stack and processed. This results in an asynchronous network request the browser asks the OS to make and await for and becomes associated with that DOM element.

While the browser waits for the OS to handle that network request, the second task in the macrotask queue, the removal of `<DropdownMenu.Portal>` and its `<form>` child, is sent to the call stack and processed.

This creates a race condition:
- If the network request is sent first, the logout will succeed (even if the response is not received/ignored).
- If the removal of `<form>` from the DOM (and the subsequent network request cancelation associated with it) succeeds first, the network request is canceled and the logout might fail, depending on whether the request has already been sent or not.

I think 😄

**Fix**
Detach the form submission from its DOM element:

```tsx
<DropdownMenu.Item asChild>
	<Form
		action="/logout"
		method="POST"
		className="rounded-b-3xl px-7 py-5 outline-none radix-highlighted:bg-muted-500"
		onClick={e => {
			e.preventDefault()
			submit(null, { action: '/logout', method: 'POST' })
		}}
	>
		<button type="submit">Logout</button>
	</Form>
</DropdownMenu.Item>
```

**Result**
- No more canceled network requests seen in Playwright tests
- No more `Form submission canceled because the form is not connected` console warnings

**Screenshots**
Network tab:
Screenshot 2023-06-15 at 10 05 40
Console:
Screenshot 2023-06-15 at 10 05 52


andrecasal:
Just noticed the logout POST request is still aborted. This only happens on Playwright tests.

```ts
//...
await page.getByRole('link', { name: user.name ?? user.username }).click()
await page.getByRole('menuitem', { name: /logout/i }).click()
await expect(page).toHaveURL(`/`) // 👈 This doesn't help

await page.goto('/login')
await expect(page).toHaveURL(`/login`)
//...
```

Screenshot 2023-06-15 at 11 02 35

app/root.tsx
```diff
@@ -235,10 +235,11 @@ function UserDropdown() {
 					</DropdownMenu.Item>
 					<DropdownMenu.Item asChild>
 						<Form
-							action="/logout"
-							method="POST"
 							className="rounded-b-3xl px-7 py-5 outline-none radix-highlighted:bg-night-500"
-							onClick={e => submit(e.currentTarget)}
+							onClick={e => {
+								e.preventDefault()
```
kentcdodds, line 240: Is there a reason this is not the only line changed in the PR? Why can't we keep the `action` and `method` and use `submit(e.currentTarget)`?

andrecasal: Let me try that out

andrecasal: I'm not sure how the Radix UI internals work, but my suspicion is that, because the form takes up all of the `<DropdownMenu.Item asChild>`'s behavior, having an onClick on it is already too late, because it'll send the close comand to the `<DropdownMenu.Portal>`.

Adding `onClick={e => e.stopPropagation()}` to the button works 👌

app/root.tsx
```diff
@@ -238,9 +236,10 @@ function UserDropdown() {
 							action="/logout"
 							method="POST"
 							className="rounded-b-3xl px-7 py-5 outline-none radix-highlighted:bg-night-500"
-							onClick={e => submit(e.currentTarget)}
```
kentcdodds, line 241: Unfortunately this change makes the form not keyboard accessible. Also, if we do this we'll want to move the padding from the form to the button so the entire row is clickable.

andrecasal:
Should be ok now 👌

✅ Detached from DOM, so no console warning
✅ Keyboard accessible
✅ Moved padding to button

kentcdodds:
I believe I've solved it another way that I'm happier with. Can you try it out and let me know if this fixes it.

I also added a logout button to the profile page because otherwise you can't logout if JS doesn't load.

kentcdodds:
Oh, I missed something: https://github.com/epicweb-dev/epic-stack/commit/a61c713f26ba9d042e70a06f0e02c95f81b6f14c

andrecasal:
Seems pretty solid 👌

kentcdodds:
Awesome!

kentcdodds:
Thank you for your help

PR #187, opened 2023-06-15
Fix username creation based on username rules

andrecasal:
This was causing flaky tests. If usernames exceeded 20 characters, the tests would fail. This fixes it.


kentcdodds (approved):
Thanks!

kentcdodds:
Looks like there's a conflict

andrecasal:
Oh, we should keep the `.slice(0, 20)`, as per the username rules.

kentcdodds (approved):
Thanks!

PR #191, opened 2023-06-16
integrate shadcn components

rajeshg:
Integrate shadcn components. To keep it simple, I tried to not change the UI much and stick with the current look and feel.
## Checklist

- [x] Tests updated
- [ ] Docs updated


kentcdodds:
Thanks for working on this! @andrecasal is working on updating the tailwind config as well with #146 so I'm concerned there will be overlap in the effort here. Maybe it would be best for me to just do this myself?

I definitely don't want to bring in components that aren't actually used in the app.

rajeshg:
Yes. Should be easy to discard the components that are not used currently. Was debating myself on starter with a bunch of handy UI components vs having only the necessary ones and leaving it up to user for the rest. 

Yes. I realized that @andrecasal  was looking into the tailwind config, so kept the changes to a minimum. This could serve as a starter for more changes. Feel free to close this, if you are looking into making these changes yourself.

kentcdodds:
This is done now #224

PR #193, opened 2023-06-17
Styling-with-css-variables-v2

andrecasal:
Added theming based on CSS Vars.

With this strategy, semantic classes (success, danger, warning, info, brand) make more sense than literal classes (red, purple, pink, etc). Because you use both literal (red, purple, etc) and semantic (day, night), and you wanted to keep them, I kept them there, but I'd suggest going for exclusively semantic classes.

Here's how I configure my own projects:
```ts
// ...
			colors: {
				background: 'rgb(var(--color-background))',
				foreground: 'rgb(var(--color-foreground))',
				brand: {
					DEFAULT: 'rgb(var(--color-brand))',
					muted: 'rgb(var(--color-brand-muted))',
				},
				success: {
					title: 'rgb(var(--color-success-title))',
					DEFAULT: 'rgb(var(--color-success-foreground))',
					background: 'rgb(var(--color-success-background))',
				},
				info: {
					title: 'rgb(var(--color-info-title))',
					DEFAULT: 'rgb(var(--color-info-foreground))',
					background: 'rgb(var(--color-info-background))',
				},
				warning: {
					title: 'rgb(var(--color-warning-title))',
					DEFAULT: 'rgb(var(--color-warning-foreground))',
					background: 'rgb(var(--color-warning-background))',
				},
				danger: {
					title: 'rgb(var(--color-danger-title))',
					DEFAULT: 'rgb(var(--color-danger-foreground))',
					background: 'rgb(var(--color-danger-background))',
				},
				'muted-50': 'rgb(var(--color-muted-50))',
				'muted-100': 'rgb(var(--color-muted-100))',
				'muted-200': 'rgb(var(--color-muted-200))',
				'muted-300': 'rgb(var(--color-muted-300))',
				'muted-400': 'rgb(var(--color-muted-400))',
				'muted-500': 'rgb(var(--color-muted-500))',
				'muted-600': 'rgb(var(--color-muted-600))',
				'muted-700': 'rgb(var(--color-muted-700))',
				'muted-800': 'rgb(var(--color-muted-800))',
				'muted-900': 'rgb(var(--color-muted-900))',
				'muted-950': 'rgb(var(--color-muted-950))',
				ring: 'rgb(var(--ring))',
			},
// ...
```

Then I just adjust these values in dark mode.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -141,12 +141,12 @@ function App() {
 				<meta name="viewport" content="width=device-width,initial-scale=1" />
 				<Links />
 			</head>
-			<body className="flex h-full flex-col justify-between bg-day-300 text-white dark:bg-night-700 ">
+			<body className="flex h-full flex-col justify-between bg-background text-foreground">
 				<header className="container mx-auto py-6">
 					<nav className="flex justify-between">
 						<Link to="/">
-							<div className="font-light text-black dark:text-white">epic</div>
-							<div className="font-bold text-black dark:text-white">notes</div>
+							<div className="font-light">epic</div>
+							<div className="font-bold">notes</div>
 						</Link>
 						<div className="flex items-center gap-10">
 							{user ? (
@@ -166,8 +166,8 @@ function App() {
 
 				<div className="container mx-auto flex justify-between">
 					<Link to="/">
-						<div className="font-light text-black dark:text-white">epic</div>
-						<div className="font-bold text-black dark:text-white">notes</div>
+						<div className="font-light">epic</div>
+						<div className="font-bold">notes</div>
 					</Link>
 					<ThemeSwitch userPreference={data.requestInfo.session.theme} />
 				</div>
@@ -197,7 +197,7 @@ function UserDropdown() {
 					to={`/users/${user.username}`}
 					// this is for progressive enhancement
 					onClick={e => e.preventDefault()}
-					className="flex items-center gap-2 rounded-full bg-night-500 py-2 pl-2 pr-4 outline-none hover:bg-night-400 focus:bg-night-400 radix-state-open:bg-night-400"
+					className="flex items-center gap-2 rounded-full bg-brand-500 py-2 pl-2 pr-4 outline-none hover:bg-brand-400 focus:bg-brand-400 radix-state-open:bg-brand-400"
 				>
 					<img
 						className="h-8 w-8 rounded-full object-cover"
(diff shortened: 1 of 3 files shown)


kentcdodds:
Let's go with semantic. Using night/day with with css-vars doesn't really make sense. Let's just adjust things to be more like what shadcn says, only update the colors to match the original design.

andrecasal:
> Let's go with semantic. Using night/day with with css-vars doesn't really make sense. Let's just adjust things to be more like what shadcn says, only update the colors to match the original design.

The only place `dark:` was being used was in the `root.tsx` to switch the `body`'s background color and the "Epic Notes" text color, which are both unnecessary now that we have `bg-background` and `text-foreground` on the `body`.

So at first glance, it would seem `-day-` and `-night-` colors would map nicely to `-muted-` colors in light and dark modes respectively, but they are being used in **both** light and dark modes - so they're unrelated to the theme.

Try to map them yourself and you'll see there's no direct pattern here 😉

kentcdodds:
I guess my point is this: I want the design to look the same as before, but I'd like to follow the pattern established by shadcn. Does that not work?

andrecasal:
It works, it's just not obvious how they should map.

Here's my attempt:
| Original | Semantic | Why |
|---|---|---|
|`-accent-purple` | `-brand-primary`| Used for primary button's background |
|`-accent-purple-lighter` | `-brand-primary-muted` | Used for secondary button's active border |
|`-accent-yellow` | `-brand-secondary`| Used for a hovered button's background |
|`-accent-yellow-muted` | `-brand-secondary-muted`| Used for an active button's background |
|`-accent-pink` | `-brand-tertiary`| Used for and checkbox's background |
|`-accent-red` | `-danger`| Used for form errors |

### Notes
- Both light and dark themes have dark backgrounds with white text, so `-night-*` classes don't map to `-muted-*` (shades of grey).
- Because the text now changes color in tandem with the theme, the hard-coded dark backgrounds have dark text too. I suggest we change the design to allow for both light and dark modes, but if you want to force the dark backgrounds in light mode, then we need to also hard-code white text on those components (inputs, profile page, notes page, dialogs, etc).
- Only the `-day-300` class is used out of the `-day-*` classes, so I mapped it to `-background`. It's not obvious what we should do with the other ones.
- `-accent-purple-lighter` is used for the secondary button's active border but it had no associated value, so I gave it a lighter `-accent-purple` (now `-brand-primary-muted`).



### Missing classes
- `-success`
- `-warning`
- `-info`
- `-muted-*`

kentcdodds:
Thank you so much for all this work!

andrecasal:
It still has dark text on dark backgrounds. I'd be happy to repaint it with light and dark modes.

kentcdodds:
I'm going to start working today so I'll fix any major issues. I'm also going to have some help from my team at EpicWeb.dev to tighten things up from a design perspective as well.

Issue #69, opened 2023-05-19
dependency check


onemen:
running `npx depcheck ` reveals some unused and missing dependencies

**Unused dependencies**
* @radix-ui/react-popover
* @radix-ui/react-separator
* @radix-ui/react-tabs
* @reach/dialog
* downshift
* remix-auth-socials
* spin-delay

**Missing dependencies**
* esbuild: .\other\build-server.ts

there are also report for **Unused devDependencies** but i don't sure how to check if those are really unused
<details>
  <summary>Unused devDependencies</summary>

* @remix-run/testing
* @testing-library/dom
* @testing-library/jest-dom
* @testing-library/react
* @testing-library/user-event
* @types/eslint
* @types/testing-library__jest-dom
* @types/uuid
* @vitest/coverage-c8
* autoprefixer
* c8
* postcss
* prettier-plugin-sql
* prettier-plugin-tailwindcss
* tailwindcss
* tailwindcss-radix
* tsconfig-paths
* vitest-mock-extende

</details>

**Outdated dependencies**
 @faker-js/faker                ^7.6.0  →   ^8.0.1 🔴
 @remix-run/router              ^1.6.1  →   ^1.6.2
 @testing-library/dom           ^9.2.0  →   ^9.3.0
 @types/cookie                  ^0.5.0  →   ^0.5.1
 @types/glob                     7.2.0  →    8.1.0
 @types/node                  ^18.16.3  →  ^20.2.1
 @vitest/coverage-c8           ^0.31.0  →  ^0.31.1
 chalk                          ^4.1.2  →   ^5.2.0 🔴
 glob                           ^7.2.3  →  ^10.2.5 🔴
 lru-cache                     ^7.18.3  →   ^9.1.1 🔴
 prettier-plugin-tailwindcss    ^0.2.8  →   ^0.3.0
 vite                           ^4.3.5  →   ^4.3.8
 vitest                        ^0.31.0  →  ^0.31.1

we can upgrade to chalk 5.2.0 now that we are using ESM
why you are still using year old glob?

onemen:
there are some breaking changes

**@faker-js/faker ^7.6.0 → ^8.0.1**
[upgrading faker to v8](https://fakerjs.dev/guide/upgrading.html)
faker deprecate [`helpers.unique` #1785](https://github.com/faker-js/faker/issues/1785) for removal in version 9.0, we need to find a replacement or write the code

**chalk ^4.1.2 → ^5.2.0**
[...]

PR #196, opened 2023-06-17
Use enforce unique instead of memoize unique

MansurAliKoroglu:
Used `enforce-unique` package for unique usernames in test utils and removed `memoize-unique` function.

https://www.npmjs.com/package/enforce-unique

This change is related with:

Issue: #69 
PR: #116 

## Test Plan

- All tests depends on `createUser` db util should pass.
- DB seeding should pass

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

No need


tests/db-utils.ts
```diff
@@ -1,16 +1,17 @@
...
+import { UniqueEnforcer } from 'enforce-unique'
 
 export function createUser() {
 	const firstName = faker.person.firstName()
 	const lastName = faker.person.lastName()
 
-	const username = unique({
-		firstName: firstName.toLowerCase(),
-		lastName: lastName.toLowerCase(),
+	const uniqueUsernameEnforcer = new UniqueEnforcer()
```
kentcdodds, line 9: If I understand the library properly, we need to create this object outside the createUser function.

kentcdodds (changes requested):
Thanks! One change please.

tests/db-utils.ts
MansurAliKoroglu: Changed

kentcdodds (approved):
Thanks!

PR #197, opened 2023-06-18
wrap table names in quotes

kashnigahbaruda:
I started getting "SqliteError: near "Order": syntax error" after migrating my db and running tests. After a long while
I think I found the 'bug' was due to my model being named 'Order' which is a SQL keyword. 

By wrapping the table names in quotes solves this issue.

This is my first PR ever so I hope I'm not doing something wrong! ... about to click the green button...


kentcdodds (approved):
Thanks!

PR #200, opened 2023-06-19
support inline usage of color-scheme CSS property

nathanvale:
https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme This CSS property is used to specify the preferred color scheme to use in the supported user agent.

In the context of this app the scroll bars now match the selected theme.

Also dropped in  `theme-color` meta support

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

## Screenshots

BEFORE
Screenshot 2023-06-19 at 7 03 44 pm

AFTER
Screenshot 2023-06-19 at 7 04 24 pm

### DIFF
--- a/app/styles/tailwind.css
+++ b/app/styles/tailwind.css
@@ -4,6 +4,7 @@
 
 @layer base {
 	:root {
+		color-scheme: light;
 		--color-background: 240 51% 91%; /* day-300 */
 		--color-foreground: 0 0% 0%; /* black */
 		--color-brand-primary: 252 100% 63%;
@@ -29,6 +30,7 @@
 	}
 
 	.dark {
+		color-scheme: dark;
 		--color-background: 0 0% 4%; /* night-700 */
 		--color-foreground: 0 0% 100%; /* white */
 	}


andrecasal:
That's super cool @nathanvale, thanks  🎉

Would it be possible for the `theme-color` meta to inherit the exact background color from the CSS Var?

Something like:
```tsx
<meta name="theme-color" media={`(prefers-color-scheme: ${theme})`} content="var(--color-background)" />
```

Also, what happens to the `colorScheme` when you set it to an invalid value like 'system'? Is it just ignored?

app/root.tsx
```diff
@@ -133,12 +133,26 @@ function App() {
 	const theme = useTheme()
 
 	return (
-		<html lang="en" className={`${theme} h-full`}>
+		<html
+			lang="en"
+			className={`${theme} h-full`}
+			style={{ colorScheme: theme }}
```
kentcdodds, line 139: I'd prefer this to be set in the css file.

app/root.tsx
```diff
@@ -133,12 +133,26 @@ function App() {
...
+		>
 			<head>
 				<ClientHintCheck nonce={nonce} />
 				<Meta />
 				<meta charSet="utf-8" />
 				<meta name="viewport" content="width=device-width,initial-scale=1" />
+				<meta
+					name="theme-color"
+					media="(prefers-color-scheme: light)"
+					content="white"
```
kentcdodds, line 149: I think we'll want these to be one of our tailwind colors. Unfortunately the value of this attribute cannot be a CSS variable so we'll either need to read it from the tailwind config at runtime in the loader and pass it into here or we'll have to hard-code it. I'm not sure which of those options I dislike least. Let's remove the theme-color for now until our design settles a bit.

kentcdodds (commented):
Thanks! I have a couple comments.

app/root.tsx
nathanvale: I looked into this yesterday and was going to get back to you with the same response. Content attribute cannot be a variable. 

I believe a solution maybe to use CSS media query. And then I cam e to the same conclusion that your css theme isn't quite there yet to super this. This is conflating anyway with the feature of this PR (`color-scheme`) so I have removed `theme-colour` support.

```
		@media (prefers-color-scheme: light) {
			meta[name="theme-color"] {
				content: var(--background);
			}
		}

		@media (prefers-color-scheme: dark) {
			meta[name="theme-color"] {
				content: var(--background);
			}
		}
```

nathanvale:
> That's super cool @nathanvale, thanks 🎉
> 
> Would it be possible for the `theme-color` meta to inherit the exact background color from the CSS Var?
> 
> Something like:
> 
> ```tsx
> <meta name="theme-color" media={`(prefers-color-scheme: ${theme})`} content="var(--color-background)" />
> ```
> 
> Also, what happens to the `colorScheme` when you set it to an invalid value like `system`? Is it just ignored?

`color-scheme:system` is not allowed but the useTheme hook only returns light or dark.

app/root.tsx
nathanvale: Done! No need for javascript at all.

A very minor change in this PR but great that we don't have white scrollbars (in my chrome browser anyway) when in dark mode!

kentcdodds (approved):
Excellent! Thank you!

PR #192, opened 2023-06-17
speed up deployments by caching playwright browsers

nathanvale:
Deployments tend to take some time (up to 8 mins at times) when `npx playwright install --with-deps` is executed. This PR implements caching for this step as NPM caching with `bahmutov/npm-install@v1`  doesn't take care of this.

## Test Plan

Merge this into a dev branch and checkout how the deployment improves timings. (Of course by running the deployment more than once.)

## Checklist

- [0 ] Tests updated
- [0] Docs 

Screenshot 2023-06-17 at 5 24 10 pm

PR #204, opened 2023-06-20
devops: do not cache Playwright browsers

mxschmitt:
Caching browsers is actually **not recommended** since the amount of time it saves, is usually the same as the amount it takes to restore it from the cache (to cache the browser binaries).

We need to understand that there are browser binaries (cacheable) and system dependencies which are installed via apt (not cacheable).

Since we are only testing with Chromium, GitHub Actions already has most of the system dependencies installed, which are required to launch Chromium. If you would test in WebKit or Firefox it would end up in errors, that the system dependencies are not there, e.g. WebKit is not able to launch, since you install the system dependencies conditionally after https://github.com/epicweb-dev/epic-stack/pull/192. So it was working more by accident.

This PR reverts https://github.com/epicweb-dev/epic-stack/pull/192, makes a pure cosmetic change to the `package.json` and then instead of installing all the browsers which we did before (npx playwright install), now we only install chromium, as well as only the chromium dependencies, this should make it already much faster. I re-use this from the package.json.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -91,21 +91,7 @@ jobs:
       - name: 📥 Download deps
         uses: bahmutov/npm-install@v1
 
-      - name: 📥 Store Playwright's Version
-        run: |
-          PLAYWRIGHT_VERSION=$(npm ls @playwright/test | grep @playwright | sed 's/.*@//')
-          echo "Playwright's Version: $PLAYWRIGHT_VERSION"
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thank you for this catch!

mxschmitt:
fyi @nathanvale, we'll also update docs according to that so its more clear: https://github.com/microsoft/playwright/pull/23814.

kentcdodds:
Thank you Max!

PR #216, opened 2023-06-24
Update cookie expiration logic for "remember me" feature

pholawat-tle:
When "remember me" is checked, rather than using a hard-coded value as `maxAge`, `expires` value should be set to session.expirationDate

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

The expiration date in the login flow is now 30 days after login date (same as onboarding flow) rather than 7 days

https://github.com/epicweb-dev/epic-stack/assets/82935610/9ad0b3e2-27d6-41ba-9889-c2f7bb133f57

### DIFF
--- a/app/routes/_auth+/onboarding.tsx
+++ b/app/routes/_auth+/onboarding.tsx
@@ -111,7 +111,7 @@ export async function action({ request }: DataFunctionArgs) {
 	cookieSession.unset(onboardingEmailSessionKey)
 
 	const newCookie = await commitSession(cookieSession, {
-		expires: remember ? session.expirationDate : undefined,
+		expires: remember ?  undefined : session.expirationDate,
 	})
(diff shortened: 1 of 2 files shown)


kentcdodds (changes requested):
I think this is actually working as designed. To make a cookie expire at the end of a browsing session, you simply don't set the expiration date.

So here's what the cookie looks like when I don't have "remember me" checked:

image

And here's what it looks like when I do have it checked:

image

When I have remember me checked, then it sets the date so even at the end of the browsing session, the user's session is remembered.

Now we do actually have a bug. The login should use the `session.expirationDate` instead of the hardcoded 7 days. Would you like to adjust this PR to fix that bug?

pholawat-tle:
I see, thank you for explaining 🙏

kentcdodds (approved):
Thank you!

PR #217, opened 2023-06-24
Adding flash storage to Epic Stack + confetti rain down & toast messages

AlemTuzlak:
## Checklist

- [ ] Tests updated
- [ ] Docs updated


package.json
```diff
@@ -74,6 +75,7 @@
     "qrcode": "^1.5.3",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-toastify": "^9.1.3",
```
rajeshg, line 78: Since we will likely we switching to using shadcn ui components in the near future (see #191), may be we can use https://ui.shadcn.com/docs/components/toast component here.

kentcdodds:
Thanks for your patience. I just pushed an update to adopt shadcn/ui: https://github.com/epicweb-dev/epic-stack/discussions/224

So now feel free to add the shadcn/ui Toast component for this and we can get this in: https://ui.shadcn.com/docs/components/toast

Thanks a lot! I think this will be a great addition!

AlemTuzlak:
@kentcdodds I think I screwed up my rebase, the changes are there but will need to create another PR probably to remove the 91 file changes listed in the PR

PR #221, opened 2023-06-27
Fix: Uppercase app names result in deployment failing

### LINKED ISSUE #218: Deployment setup fails if project name has capital letters
When running `npx create-remix@latest --typescript --install --template epicweb-dev/epic-stack`
You get asked `Where would you like to create your app?`, I entered `HousingHub`, from how I know it from other frameworks like Next, you usually get an error there to only use lower case.
Here however you don't and fly doesn't like that.
image
Adding a .toLowerCase() to the APP_NAME in [remix.init](https://github.com/epicweb-dev/epic-stack/tree/main/remix.init/index.mjs) line 32, probably does the trick here.

NotLuksus:
Closes #218


kentcdodds (approved):
Thank you!

PR #230, opened 2023-06-29
feat: devcontainer configuration

andrewbrey:
This PR adds `devcontainer` configuration files to the `Epic Stack`.

You can find a lot more info about `devcontainers` on the [VS Code Docs](https://code.visualstudio.com/docs/devcontainers/containers) page about them as well as on the [devcontainer specification](https://containers.dev/), but the `tl;dr` is that this set of configurations allows Epic Stack projects to leverage remote development environments such as **GitHub Codespaces**, **Gitpod**, **VS Code Remote Containers** (machine-local docker container), and a growing list of other supported providers.

This is the underpinning of how GitHub themselves [does development on the GitHub web application](https://github.blog/2021-08-11-githubs-engineering-team-moved-codespaces/) and it's an incredible value-add, especially for open source repositories because it can lower the barrier to entry significantly for new contributors by getting rid of so many potential stumbling blocks related to environment setup.

For the Epic Stack, the configurations leverage the existing multi-stage docker build setup found in `other/Dockerfile` so that the development environment provided to developers is _nearly identical_ to that which will run in production (with the minor difference that the devcontainer also adds `git` and `sudo` which are utilities not present in the production image.

## Test Plan

Spin up your own Codespace in GitHub against this branch, and after a few minutes of bootstrapping (which only happens once at Codespace creation but doesn't happen on subsequent re-starts of the Codespace) you are plopped into the web version of VS Code and everything works just like you were developing locally, except your code will be running on GitHub servers and your dev environment will spin up against real-looking GitHub subdomains instead of "localhost" urls!

Now anybody can code on their Epic Stack application from any device that can run VS Code in the browser (yes, it works on phones)!

## Checklist

- [ ] Tests updated
- [ ] Docs updated

Tests are not really applicable to this change.

## Screenshots

**Demo in VS Code (_running locally with the "remote" environment being a docker container_)**
https://youtu.be/3R8H_AMZ6PI

**Demo in GitHub Codespaces (_running in the cloud on GitHub hardware_)**
https://youtu.be/XP_y0w0wWJk

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json.
+{
+  "name": "app",
+  "dockerComposeFile": "docker-compose.yml",
+  "service": "app",
+  "workspaceFolder": "/workspace",
+  "remoteUser": "node",
+  "postCreateCommand": [".devcontainer/post-create.sh"],
+  "containerEnv": {
+    "NODE_ENV": "development"
+  },
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {},
+    "ghcr.io/devcontainers-contrib/features/apt-get-packages:1": {
+      "packages": "sudo"
+    },
+    "ghcr.io/wxw-matt/devcontainer-features/command_runner:latest": {
+      "command1": "echo 'node ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers.d/node"
+    }
+  },
+  "customizations": {
+    "vscode": {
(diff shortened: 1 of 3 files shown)


imcodingideas:
Thank you.

kentcdodds:
Thanks for this! I'd welcome [an example](https://github.com/epicweb-dev/epic-stack/blob/main/docs/examples.md), but I don't think that I will include this as part of the starter. Would you be interested in converting this to a new PR that makes an example?

andrewbrey:
No worries, I'll go ahead and close out this PR  - cheers!

PR #238, opened 2023-06-29
Flash session storage added with confetti and toast notifications

kentcdodds:
I had a few adjustments to make to #236 but couldn't push directly to that branch, so here we are :)

Closes #236


kentcdodds:
Thanks @AlemTuzlak!

PR #243, opened 2023-06-30
add tailwind class sorting

andrecasal:
This plugin is already installed in package.json, but I don't think we're using it.

This change add tailwind class sorting.

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -25,4 +25,5 @@ module.exports = {
 			},
 		},
 	],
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
I don't think it's required to be included here, but it makes sense to include so let's go for it :)

MichaelDeBoey:
Normally it should have been picked up automatically, but no harm in explicitly enabling it either

andrecasal:
Hum, class sorting was disabled for me until I explicitly added the plugin. I wonder why 🤔

PR #244, opened 2023-06-30
docs(deployment): connect with prisma studio to deployed app

musyyy:
Just updated the documentation based on a [discord thread](https://discord.com/channels/715220730605731931/1123737664814854164/1123835430731776112).

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -106,6 +106,22 @@ command-line session on the machine. You can do this using `fly ssh console`.
 The Dockerfile simplifies this further by adding a `database-cli` command. You
 can connect to the live database by running `fly ssh console -C database-cli`.
 
+To connect to the deployed database from your local machine using Prisma Studio, 
+you can utilize Fly's ﻿`ssh` and ﻿`proxy` commands.
+
+- Run in one terminal the command to start Prisma Studio on your desired Fly app
+  ```sh
+  fly ssh console -C "npm run prisma:studio" --app [YOUR_APP_NAME]
+  ```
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -106,6 +106,22 @@ command-line session on the machine. You can do this using `fly ssh console`.
...
+you can utilize Fly's ﻿`ssh` and ﻿`proxy` commands.
+
+- Run in one terminal the command to start Prisma Studio on your desired Fly app
+  ```sh
+  fly ssh console -C "npm run prisma:studio" --app [YOUR_APP_NAME]
+  ```
+- Run in a second terminal the command to proxy your local port 5556 to 
+  Prisma Studio
+  ```sh
+  fly proxy 5556:5555 --app [YOUR_APP_NAME]
```
kentcdodds, line 119: Any reason this isn't `5555:5555`? Is it just to help differentiate prod from dev? If so, let's make a note about that.

kentcdodds (commented):
This looks great. Just one question.

docs/deployment.md
musyyy: In general I like to use a different port to explicitly imply which one the local port is, and maybe you might run your local prisma studio.

musyyy: Do you think it would be a good DX define a `package.json` `script` like `npm run prisma:studio:stg` or `npm run prisma:studio:prod`?

kentcdodds: > to explicitly imply which one the local port is

That's a great idea 👍

I don't think we need the scripts in the package.json. I think we're good here 👍

PR #246, opened 2023-06-30
Make modal dialog exit animations visible

benhsm:
Problem: the `/settings/profile` route uses an `<Outlet />` to display modals from child routes `photo` and `two-factor`. The models are opened and closed by client-side navigation to the child routes and back. The shadcn ui dialog component comes with CSS entry and exit animations, but when the navigation happens to exit the modal, it cuts off the animation before it can complete.

I drew on a blog post from Jacob Paris here for my approach to addressing this: <https://www.jacobparis.com/content/remix-animated-page-transitions>

Summary of changes:
- Wrap the `useOutlet` hook with `useState` and replace the `<Outlet />` with it
- Add SwitchTransition and CSSTransition from react-transition-group
- Update modal component's `useLoaderData` hooks to be able to fallback to using an empty object so that components don't error out when data is unloaded by navigation
- Control whether the modal dialogs are open or closed using `useState` and trigger them to close before navigation happens

## Test Plan

- Verify that the e2e tests related to profile photos and 2fa still pass
- Manually check that profile photo and 2fa modals look ok when closed:
     - by clicking outside of the modal
     - by clicking the 'x' icon
     - by submitting the form 

## Checklist

- [x] Tests updated: 
    - add a line to the profile photo settings e2e test so that playwright checks that the dialog is hidden before it checks that the redirect happens
    - change 2fa test to close modal dialog using 'x' icon as a button rather than a link 
- No docs changed.

## Screenshots

Before: before

After: after

I think these changes are an improvement, but I'd certainly appreciate oversight and review here. I'm a little unsure about the case where the `profile.two-factor.verify` modal is closed by form submission; it is possible that that may warrant further work.


app/routes/settings+/profile.photo.tsx
```diff
@@ -211,14 +218,13 @@ export default function PhotoChooserModal() {
 					<ErrorList errors={form.errors} />
 				</Form>
 				<DialogClose asChild>
-					<Link
-						to=".."
-						preventScrollReset
+					<button
```
kentcdodds, line 221: I would definitely prefer to keep this a link because we are navigating. Is there a reason this was changed for a button?

benhsm: For the exit animation to display, the state of the dialogs `open` prop has to be changed to closed in close time-proximity to when the navigation is triggered, and this should happen the same way whether the cause of the exit is `onEscapeKeyDown`, `onPointerDownOutside` or clicking this `Icon` component. When I was writing this, it seemed easiest and most consistent to handle all of those events with the same `dismissModal` function that changed the state and then triggered navigation, and that navigation was redundant with the navigation triggered by the link, so I changed it to a button.

However, the following also seems to work:
```js
<DialogClose asChild>
	<Link
		to=".."
                onClick={() => setOpen(false)}
		preventScrollReset
		aria-label="Close"
		className="absolute right-20 top-20"
	>
		<Icon name="cross-1" />
	</Link>
</DialogClose>
```

Should I make this change?

kentcdodds: Yes. The benefit of this is that people can use it like a link (like right clicking and opening in a new tab etc). So I *always* put in the extra work to make all user-initiated navigations use a Link.

While you're at it, could you rebase on `main` and run `npm run format`?

benhsm: Sure thing

benhsm:
I'll have to see what's going on with that test later today

tests/e2e/2fa.test.ts
```diff
@@ -35,7 +35,7 @@ test('Users can add 2FA to their account and use it when logging in', async ({
 		dialog.getByRole('button', { name: /disable 2fa/i }),
 	).toBeVisible()
 
-	await dialog.getByRole('link', { name: /close/i }).click()
+	await dialog.getByRole('button', { name: /close/i }).click()
```
kentcdodds, line 38: I think this is the problem:

```suggestion
	await dialog.getByRole('link', { name: /close/i }).click()
```

benhsm:
I'm not sure if I'll be able to find a good way to prevent that, but I'll try to figure something out

PR #251, opened 2023-07-02
Cleanup: Update server/index.ts to use MODE

L-Steinmacher:
MODE was declared but not used in a few cases.


kentcdodds (approved):
Thanks!

PR #253, opened 2023-07-03
(examples): Add epic-stack-with-storybook

moishinetzer:
This example showcases storybook usage with the epic-stack. If there is interest to integrate this in the epic-stack I would be more than happy to expand the example and add chromatic tests and add those as part of the CI/CD pipeline.

The shadcn/ui components don't have stories as those are fairly trivial and follow the same pattern as shown in the other stories already implemented.

@kentcdodds I really do believe there is a very strong point to add storybook to the main epic-stack repo, so I would love to hear what your thoughts are after seeing this.


kentcdodds (approved):
Super! Thank you for taking the time to put this together 👏👏

PR #254, opened 2023-07-03
Update examples.md with Socket.IO example

L-Steinmacher:
Example app has a simple "ping" button that you may check the networks tab to see the socket connection.  Tests written with Vitest to show the socket is working as well.

## Test Plan
Tests written in `app/uitls/socket.test.ts`  verifies the functionality of a Socket.IO server and client using Vitest. Specifically, it checks two key features: first, it ensures that the server can successfully send a 'hello' event to the client with the payload 'world'. Second, it tests the server's acknowledgement ('ack') feature by having the client emit a 'hi' event and checking if the server correctly responds with 'hola'.

## Checklist

- [x] Tests updated
- [x] Docs updated


kentcdodds (approved):
Awesome! Thanks for working on this and adding the example 👏👏

Issue #257, opened 2023-07-03
2FA not working with Google Authenticator?


Mat-moran:
Hi,

I was testing the 2FA authentication flow when I realised that is not working with Google Authenticator App. With other providers works well.

## Steps to reproduce
- Create a brand new epic stack app
- Start the development server
- Create an account
- Try to enable 2FA reading the QR code with google Authenticator app (I am using version 4.0.1 of this app in IOS 16.5.1)
- You should get an error saying that the code is not valid.

I have tried in IOS and in Android without success.

amir-ziaei:
Just tried it out on my end and it seems like the issue is within the QR code. Because it works when I enter the `secret` manually into Google Authenticator.

More, I also tried the manual flow with 1Password and everything seemed fine. Except, 1Password doesn't seem to be able to recognize there is a QR Code on the page to offer its auto-scanning feature. Maybe it needs some hints? Further investigation might be nice.

mediabakery:
I found the issue. Contrary to what the documentation suggests, google authenticator evaluates the algorithm field and only accepts uppercase.

kentcdodds:
Merged. Thank you very much!

PR #265, opened 2023-07-05
fix(totp): algorithm name changed to uppercase

mediabakery:
to support google authenticator without breaking other totp verifier. see #257 

Unlike other authenticator apps, google authenticator only allows the algorithms listed in the [documentation](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#algorithm). That is `SHA1`, instead of `sha1`. 

> Currently, the algorithm parameter is ignored by the Google Authenticator implementations.

Contrary to what the documentation suggests, this value is read and used!

It is very difficult to say which spelling is the correct one. If you look at the [node.js crypto documentation](https://nodejs.org/api/crypto.html#cryptocreatehashalgorithm-options), the lower case is preferred in any case. 

The function `crypto.getHashes()` returns here also only the lower case.

But in the source code of openssl you can find both variants. An alias ensures that both notations are valid. 

I added a test `Generating and verifying also works with the algorithm name alias` to ensure that both notations produce the same result.

To create consistency, I have adapted the algorithm name even with other usage e.g. forgot-password.

## Test Plan

- [x] Test openssl behaviour with different algorithm names
      ✅ `openssl dgst -SHA1`
      ✅ `openssl dgst -sha1`
      ❌ `openssl dgst -ShA1`
- [x] Test with google authenticator
- [x] Test with Microsoft Authenticator
- [ ] Test with Authy

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/routes/_auth+/forgot-password/index.tsx
+++ b/app/routes/_auth+/forgot-password/index.tsx
@@ -70,7 +70,7 @@ export async function action({ request }: DataFunctionArgs) {
 		// don't leak the association between the two.
 		const target = usernameOrEmail
 		const { otp, secret, algorithm, period, digits } = generateTOTP({
-			algorithm: 'sha256',
+			algorithm: 'SHA256',
 			period: tenMinutesInSeconds,
 		})
 		// delete old verifications. Users should not have more than one verification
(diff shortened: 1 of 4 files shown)


kentcdodds:
I have confirmed this works with 1Password. I don't use Authy, but I'm guessing if it works with Google Authenticator, 1Password, and Microsoft Authenticator it's going to work with Authy.

Thank you very much for figuring all this out!

Issue #269, opened 2023-07-05
Use email address instead of unknown user ID as totp label


mediabakery:
Currently, the user ID is displayed as the label in the Authenticator apps. However, this is unknown for the user. If, as is often the case, the unique email address is displayed instead, the user can more easily distinguish between individual accounts.

current state:
`url:userID`

Preferred state:
`url:email`

kentcdodds:
Merged. Thanks!

PR #270, opened 2023-07-05
feat(totp): Use email instead of ID in the label

mediabakery:
#269
Make the request in the Authenticator app more meaningful by displaying the email address instead of `URL:userID` (not really known to the user). `URL:email`

## Test Plan

- [ ] Scan QR Code and see your email address as label.

## Checklist

- [x] Tests updated << n.a.
- [x] Docs updated << n.a.


kentcdodds (approved):
I agree with this change. I wasn't certain where that label is displayed in other apps, but it definitely sounds more sensible to use the email. Thanks!

PR #275, opened 2023-07-06
Site manifest cross-origin fix

xHomu:
Site manifest is triggering a second SSL handshake:

## Test Plan

This affects a browser behavior, so I don't know how to test for it.

## Screenshots

image

### Cause

> If the manifest requires credentials to fetch, the [crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) attribute must be set to use-credentials, even if the manifest file is in the same origin as the current page.

https://developer.mozilla.org/en-US/docs/Web/Manifest#deploying_a_manifest

### Post-fix

image

`as const` is necessary due to weird way typescript interacts with `[].filter(Boolean)`.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -61,7 +61,11 @@ export const links: LinksFunction = () => {
 			href: '/favicons/favicon-32x32.png',
 		},
 		{ rel: 'apple-touch-icon', href: '/favicons/apple-touch-icon.png' },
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
That's weird. But this is why we have starters like this. So we can handle all the weird stuff and start people on the right foot :) Thanks a lot for the help!

Issue #274, opened 2023-07-05
Nice scrolling behavior on notes list


kentcdodds:
This is described in this twitter thread: https://twitter.com/kentcdodds/status/1676730931170910208

Check out the loom: https://www.loom.com/share/407a116d969145a3be721b2604c880ca

The relevant file is here: https://github.com/epicweb-dev/epic-stack/blob/f3459e75224aaf7c13ea0a71859070528313c714/app/routes/users%2B/%24username_%2B/notes.tsx#L84

brandonmcconnell:
@kentcdodds Here is a solution to your question that only makes a few minor changes, retains your `flex-1` wrapper, and implements the desired functionality: https://play.tailwindcss.com/cKjBnmYyLk?size=1592x645

kentcdodds:
Thanks for the help @brandonmcconnell. If you think you can improve on the solution I built with help from @im-ironclad, then please do. I'm pretty happy with the solution we have now.

PR #276, opened 2023-07-06
Improve notes scrolling

brandonmcconnell:
This…

* reverts most of the UI changes to their simpler state when @kentcdodds posted on Twitter (prior to the implemented fix from #274)
* only required a few changes
* **_notable difference_** — the scrolling from this fix allows the entire wrapper to scroll the scrollable areas rather than having smaller scrollable boxes nested in non-scrollable wrappers, which feels a bit more natural


kentcdodds (approved):
Yeah, this is definitely an improvement. Thanks!

kentcdodds:
Unfortunately this doesn't look great with long notes. Any ideas?

image


I think I'd rather make it so the note itself scrolls but the edit and delete button remains on the bottom. And I only want the note itself to scroll if there's not enough room on the page to fit it without scrolling the whole page.

PR #279, opened 2023-07-06
Improve single note content scrolling

brandonmcconnell:
See attached screen recording, below

<details><summary>old video</summary>
<p>

https://github.com/epicweb-dev/epic-stack/assets/5913254/e900b41e-96b0-4842-9a61-a8c3f93626fc

</p>
</details>

<details open><summary> new video (from <a href="#issuecomment-1623286043">below comment</a>)

</summary>
<p>

https://github.com/epicweb-dev/epic-stack/assets/5913254/e51e15c2-5881-40ec-a019-66e2d6f4e865

</p>
</details> 

Two things I also noticed that may be worth addressing separately:
* longer names get truncated without an ellipsis on smaller viewports (seen in video above)
* if you don't use the code quick enough to auth a new account, the code seems to expire, but no relevant expiration message shows when attempting to use the code except "invalid code" (or something similar)


brandonmcconnell:
Also, I went with a sticky bottom bar for the buttons on the viewing (non-editing) view, but I don't really like the way the buttons change vertical position between that view and the editing view.

Pushing a change for that shortly as well

brandonmcconnell:
Finessed the positioning and behavior a bit more for those bottom sticky areas. They still reposition slightly between those two views, but I think it feels more purposeful now, where one feels like a toolbar, whereas the other feels like a traditional form footer.

https://github.com/epicweb-dev/epic-stack/assets/5913254/e51e15c2-5881-40ec-a019-66e2d6f4e865

I also…
* added some responsive styles to those footer buttons to make them more app-like when on smaller viewports
* added a datetime to fill some excessive whitespace on desktop/tablet in that bottom toolbar area

kentcdodds:
I like these changes. Do you think we could also make it so the edit view does the same thing so the two buttons maintain the same position in the two views?

kentcdodds:
Oh, also, the icon.svg file is generated so your changes will get overridden. Check the icon docs to see how to add icons. We should probably not commit the generated file to avoid this problem, but let's do that in another PR.

brandonmcconnell:
@kentcdodds All changes done — _screen recording below_

** I also had an idea regarding the form. How would you feel about—instead of having a separate and different-looking view for the form-editing experience—using a package like https://github.com/FormidableLabs/use-editable and using essentially the same UI for the form that you already do for the view, but with editable sections, and dynamic placeholders, so not only will the bottom floating toolbars transition seamlessly, since they both use the same styles, but the content itself would appear almost identical, as the rendered elements will be largely the same.

That package I linked above appears to do a good job of managing the state for editable elements and resolves many of the usual gotchas you encounter when working with `contenteditable`.

Definitely something for a different PR, but thought it might be a nice touch 🙂

https://github.com/epicweb-dev/epic-stack/assets/5913254/f685eab1-d20c-4a36-83de-e059015c5480

brandonmcconnell:
I see there are a couple of TS bugs. I'll resolve those later.

app/components/forms.tsx
```diff
@@ -80,9 +82,11 @@ export function TextareaField({
...
 				{...textareaProps}
 			/>
-			<div className="px-4 pb-3 pt-1">
-				{errorId ? <ErrorList id={errorId} errors={errors} /> : null}
-			</div>
+			{errorId ? (
+				<div className="px-4 pb-3 pt-1">
+					<ErrorList id={errorId} errors={errors} />
+				</div>
+				) : null}
```
kentcdodds, line 89: Could you show how this impacts the UI? I'm worried that this will result in jerkiness when error messages appear. I believe that's why we have the divs always rendered.

app/routes/users+/$username_+/notes.$noteId.tsx
```diff
@@ -27,22 +30,35 @@ export async function loader({ request, params }: DataFunctionArgs) {
 
 export default function NoteRoute() {
 	const data = useLoaderData<typeof loader>()
+	const date = new Date(data.note.updatedAt)
+	const timeAgo = formatDistanceToNow(date)
```
kentcdodds, line 34: We need to do this computation in the `loader` to avoid timezone issues.

tailwind.config.ts
```diff
@@ -120,5 +121,15 @@ export default {
...
+	plugins: [
+		animatePlugin,
+		radixPlugin,
+		plugin(function ({ addComponents }) {
+			addComponents({
+				'.floating-toolbar': {
+					'@apply flex items-center absolute left-3 right-3 bottom-3 p-4 pl-5 md:pl-7 rounded-lg shadow-accent shadow-xl gap-2 md:gap-4 bg-muted/80 backdrop-blur-sm': {},
+				},
+			})
+		}),
```
kentcdodds, line 133: I would much rather put this in a new file called `app/styles/global.css` which is imported as a [css side-effect import](https://remix.run/docs/en/1.18.1/guides/styling#css-side-effect-imports) and uses [CSS bundling](https://remix.run/docs/en/1.18.1/guides/styling#css-bundling)

kentcdodds (commented):
This is looking awesome. Just a few thoughts.

app/components/forms.tsx
brandonmcconnell: Sure — by always rendering, the empty div still took up some space which could cause some weird effects when working with displays types like grid and flex

Lmk if you'd like me to add some supporting screenshots as well

app/routes/users+/$username_+/notes.$noteId.tsx
brandonmcconnell: Would you mind clarifying? The datetime I'm referencing here _is_ set in the loader and includes the timezone info, so _I think_ it should always show the accurate relative time, respective to the user's current local time.

I could be missing something here though, so please let me know.

tailwind.config.ts
brandonmcconnell: Gotcha! That works. I try to stay in "tailwind world" as much as I can, but I can how a side-loaded CSS file that might be simpler to manage and better-scoped, since you don't need it everywhere.

kentcdodds: I've made some changes of my own. I'm going to push and merge :D

kentcdodds:
https://github.com/epicweb-dev/epic-stack/assets/1500684/2390e01d-2633-4461-9471-24b2db9a2b53

tailwind.config.ts
brandonmcconnell: Awesome, sounds good!

PR #286, opened 2023-07-07
tests: make username lowercase

cezarneaga:
I was having troubles in CI with onboarding tests failing.

I managed to get them to pass by making username lowercase. i think this may have been forgotten in 53919cc5210c14584deea0451c156735584de4dc 

also i removed the regex because it was creating potentially problematic emails:
```
  name: 'Gerard Connelly',
  username: '_erard__onnelly',
  email: '_erard__onnelly@example.com',
```

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/tests/e2e/onboarding.test.ts
+++ b/tests/e2e/onboarding.test.ts
@@ -7,6 +7,7 @@ import {
 	test,
 } from '../playwright-utils.ts'
 import { readEmail } from '../mocks/utils.ts'
+import { createUser } from 'tests/db-utils.ts'
 
 const urlRegex = /(?<url>https?:\/\/[^\s$.?#].[^\s]*)/
 function extractUrl(text: string) {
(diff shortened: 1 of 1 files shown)


tests/e2e/onboarding.test.ts
```diff
@@ -20,7 +20,7 @@ test('onboarding with link', async ({ page }) => {
 	const username = faker.internet
 		.userName({ firstName, lastName })
 		.slice(0, 20)
-		.replace(/[^a-z0-9_]/g, '_')
+		.toLowerCase()
```
kentcdodds, line 23: Would this solve the weird email problem?

```suggestion
		.trim()
		.replace(/[^a-z0-9_]/g, '_')
		.toLowerCase()
```

kentcdodds (commented):
Thanks. I think it would make sense to make a function in this module for creating onboarding data.

PR #290, opened 2023-07-09
Dialog overflows when the content does not fit

abereghici:
The dialog component overflows when the content doesn't fit in the viewport.
Reproduced on a 14-inch MacBook.

## Screenshots
Before fix
Screenshot 2023-07-09 at 14 18 06

After fix
Screenshot 2023-07-09 at 14 18 58


kentcdodds (approved):
Thank you!

PR #292, opened 2023-07-10
Add meta tag to fix certain iOS hyration issues

xHomu:
> iOS attempts to detect phone numbers, email addresses, and other data in text content and convert them into links, leading to hydration mismatches.

https://nextjs.org/docs/messages/react-hydration-error#common-ios-issues


kentcdodds:
Hmmm... I think I'd rather have documentation for this situation and recommend `suppressHydrationWarning` instead. It seems a shame to disable what is probably a pretty useful feature of iOS 🤔

xHomu:
This is more of an upstream issue. Maybe an update somewhere on Remix Docs?

PR #294, opened 2023-07-10
Add time zone client hint example

kiliman:
Added link and description to the time zone client hint example.


kentcdodds (approved):
Thanks!

Issue #295, opened 2023-07-10
Support progressive enhancement with a server rendered modal


kentcdodds:
One issue with the modal is that it requires JavaScript. This means if someone lands on a route that's a modal they won't get that modal to show up until the JavaScript loads which means it could take a while to show up which would be confusing for users.

I think if we land on the page we should server render the modal without the animation. I remember doing something like this in a live stream here: https://www.youtube.com/watch?v=4k4rT31aGo0 (most interesting stuff is in the last half hour where we figure out how to do it well finally).

PR #300, opened 2023-07-11
Fix image unloading flash in modal exit animation

benhsm:
In the current implementation of modal exit animations in the `/settings/profile` route, images in the modals disappear or change to defaults during transition, producing an unpleasant 'flash' (as described and shown here: https://github.com/epicweb-dev/epic-stack/pull/246#issuecomment-1629266271).

This happens because the loader data used to derive the images is unloaded during client-side navigation from the modal's child route back to the parent route. A simple fix for this is to track the data used to derive the images as React state so that it is not affected by data unloading during navigation. This PR applies that fix to the user's profile image in `profile.photo.tsx` and the QR code image in `profile.two-factor.verify.ts`.

## Test Plan

- Check that the profile photo modal looks as expected when:
    - exiting the modal without changing the image
    - exiting the modal after deleting the image
    - exiting the modal after updating the image
- Check that the two factor verification modal looks as expected when verifying or cancelling

## Checklist

- No changes to tests or docs

## Screenshots (slowed down to make the change visible)

Before:
before

After:
after

Hopefully this won't somehow introduce complications to work on #295 .

### DIFF
--- a/app/routes/settings+/profile.photo.tsx
+++ b/app/routes/settings+/profile.photo.tsx
@@ -126,6 +126,7 @@ export async function action({ request }: DataFunctionArgs) {
 
 export default function PhotoChooserModal() {
 	const data = useLoaderData<typeof loader>() || {}
+	const [userImgId] = useState(data.user?.imageId);
 	const [newImageSrc, setNewImageSrc] = useState<string | null>(null)
 	const navigate = useNavigate()
 	const deleteImageFetcher = useFetcher<typeof deleteImageRoute.action>()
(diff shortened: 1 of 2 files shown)


benhsm:
Doing that in a simple way seems to work for removing the flash, but I'm not sure how to handle propogating changes, like an updated user profile image, from the child component back up into the parent wrapper. My mental models of react hooks and state might just be too wobbly/fuzzy here. I'll spend some time researching and considering further

kentcdodds:
Thanks for this! As we discussed on office hours, I've removed the dialog altogether: https://github.com/epicweb-dev/epic-stack/discussions/309

PR #303, opened 2023-07-12
Add timezone client-hints

rajeshg:
Full credits to @kiliman for this work. https://github.com/kiliman/epic-stack-time-zone

Client hints for timezone are a nice add. The approach is similar to the way themes are handled currently. CH-time-zone will be added to the cookies. 

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/COMMUNITY.md
+++ b/COMMUNITY.md
@@ -1,0 +1,10 @@
+# Community
+
+Here you can find useful learning resources and tools built and maintained by the community, such as libraries, examples, articles, and videos. 
+
+## Learning resources
+
+### Videos
+
+- **Dark Mode Toggling using Client-preference cookies** by [@rajeshdavidbabu](https://github.com/rajeshdavidbabu)  - Youtube [link](https://www.youtube.com/watch?v=UND-kib_iw4)
+


app/routes/users+/$username_+/notes.$noteId.tsx
```diff
@@ -112,7 +115,7 @@ export default function NoteRoute() {
 			{data.isOwner ? (
 				<div className={floatingToolbarClassName}>
 					<span
-						className="text-sm text-foreground/90 max-[524px]:hidden"
+						className="inline-flex text-sm text-foreground/90 max-[524px]:hidden"
```
rajeshg, line 118: Fixes a minor alignment issue with the clock icon and the text that follows

app/utils/client-hints.tsx
```diff
@@ -15,6 +15,14 @@ export const clientHints = {
 			return value === 'dark' ? 'dark' : 'light'
 		},
 	},
+	timeZone: {
+		cookieName: 'CH-time-zone',
+		getValueCode: `Intl.DateTimeFormat().resolvedOptions().timeZone`,
+		fallback: 'UTC',
+		transform(value: string | null) {
+			return value ?? 'UTC'
+		},
```
kentcdodds, line 24: In this case, we don't need to transform the value at all and the fallback is already defined. I think we should not require the `transform` function. So let's remove it here.

```suggestion
```

This change may necessitate a small update in the code below.

kentcdodds (changes requested):
I love this. Just a few changes.

app/root.tsx
```diff
@@ -106,6 +111,13 @@ export async function loader({ request }: DataFunctionArgs) {
 		await authenticator.logout(request, { redirectTo: '/' })
 	}
 	const { flash, headers: flashHeaders } = await getFlashSession(request)
+	// default options to use timeZone client hint
+	const localDateTimeFormat = getDateTimeFormat(request)
+	const serverDateTimeFormat = getDateTimeFormat(request, {
+		timeZone: 'UTC', // override timezone here
+		// month: 'short', // override month style, etc.
+	})
+	const now = new Date()
```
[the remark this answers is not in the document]
kiliman: Yeah, that code was in my example to show the difference between server and local time. Definitely not meant to be part of the default template.

app/utils/client-hints.tsx
```diff
@@ -53,19 +64,20 @@ export function getHints(request?: Request) {
...
+			acc[hintName] = hint.transform
+				? hint.transform(getCookieValue(cookieString, hintName))
+				: hint.fallback
 			return acc
 		},
 		{} as {
-			[name in ClientHintNames]: ReturnType<
-				(typeof clientHints)[name]['transform']
-			>
+			[name in ClientHintNames]: string
```
rajeshg, line 73: We lose some type-checking here. But couldn't think of a better way to do it.

kentcdodds: Ah yeah, we'll need to figure out how to have our cake and eat it too here 😅 If you can't work it out, then that's fine. I'll come back to this later when I find time. Thanks!

rajeshg: Couldn't figure out :(

kentcdodds:
Why was this PR closed? I'd still like to have this feature eventually and I'd prefer you get git credit for your work :)

I might recommend you make any future PRs from a non-`main` branch though.

rajeshg:
My mistake. You are right, I should have done that from a non-main branch. I attempted to copy over the latest changes from epicweb-dev/epic-stack to rajeshg/epic-stack.

rajeshg:
Reopening this again

rajeshg:
Couldn't reopen this myself, since I lack the rights to reopen pull requests on this repo. I will submit a separate pull request, when I get a chance.

kentcdodds:
I can't reopen it either because you created it from `main` and now your `main` branch is different. This is another reason using a different branch helps 😅 Thanks!

PR #305, opened 2023-07-13
Validate uniqueness of username before creating a user through signup

benhsm:
In the current build, if a new user going through the sign in process on the onboarding page supplies a username that is already in use, the onboarding action attempts to create a new user, violating the Prisma schema's uniqueness constraint. This throws an exception that goes all the way to the main error boundary.

This PR changes the behavior of the onboarding page so that submitting a non-unique username is treated as a form validation error instead, using the same kind of check used in `_auth+/signup/index.tsx` to verify the uniqueness of emails.

## Test Plan

- Check that the UX of the onboarding experience is as expected when onboarding with duplicate and non-duplicate usernames

## Checklist

No changes to tests or docs

## Screenshots

Before:
before

After:
after

### DIFF
--- a/app/routes/_auth+/onboarding.tsx
+++ b/app/routes/_auth+/onboarding.tsx
@@ -28,6 +28,7 @@ import {
 } from '~/utils/user-validation.ts'
 import { checkboxSchema } from '~/utils/zod-extensions.ts'
 import { redirectWithConfetti } from '~/utils/flash-session.server.ts'
+import { prisma } from '~/utils/db.server.ts'
 
 export const onboardingEmailSessionKey = 'onboardingEmail'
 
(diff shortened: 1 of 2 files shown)


app/routes/_auth+/onboarding.tsx
```diff
@@ -81,9 +82,25 @@ export async function action({ request }: DataFunctionArgs) {
 	}
 
 	const formData = await request.formData()
-	const submission = parse(formData, {
-		schema: onboardingFormSchema,
+	const submission = await parse(formData, {
+		schema: () => {
+			return onboardingFormSchema.superRefine(async (data, ctx) => {
```
kentcdodds, line 87: Could this be:

```suggestion
		schema: onboardingFormSchema.superRefine(async (data, ctx) => {
```

I'm not sure of the benefit of using a callback here.

kentcdodds (commented):
Ah thank you! This is definitely an improvement 👍 Just one idea.

app/routes/_auth+/onboarding.tsx
kentcdodds: @edmundhung is there a reason using the callback is better?

kentcdodds (approved):
Let's merge this and we'll revert if Edmund says there's a reason we shouldn't do it this way.

app/routes/_auth+/onboarding.tsx
edmundhung: It is needed only when you are setting up async live validation. This is likely a leftover from the previous implementation.

kentcdodds: > async live validation

What does this mean?

edmundhung: Async validation on input or blur event

kentcdodds: I don't understand why that would need the callback form on the server.

edmundhung: It's an advanced feature. The callback is used to retrieve the `submission.intent` and adjust the zod schema behavior accordingly. In the case of async validation on input / blur, this allows us to run async validation only when it is intent to.

You can find an example [here](https://conform.guide/validation#skipping-validation)

kentcdodds: Oh I see! Thank you for explaining that. I'm really amazed of all the things you've thought of with this!

PR #312, opened 2023-07-16
Replace code verification input with pin input component

SomiDivian:
Replace code verification input with pin input component

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (changes requested):
Thanks for this! It's definitely something I've wanted.

I think there are a lot of things the component can do that we're not using. I'd like those features removed to make it easier to maintain. It may also be worth adding a component test since this is a fairly complicated component.

app/routes/_auth+/signup_.verify.tsx
```diff
@@ -148,16 +150,15 @@ export default function SignupVerifyRoute() {
 						...conform.input(fields.email),
 					}}
 					errors={fields.email.errors}
+					className='hidden'
 				/>
 				<Field
```
zolrath, line 155: You're importing PinField above but still using a Field here?

PR #314, opened 2023-07-17
Explain cookie expires

jacobparis:
If you assume a cookie with no `expires` field doesn't expire, this remember me checkbox appears to be backwards, so I've added a comment explaining why it does work


kentcdodds (approved):
Thanks! I've definitely been asked about this.

PR #319, opened 2023-07-18
Add search e2e test and updated packages

rajeshg:
- e2e test for search
- update packages

## Checklist

- [x] Tests updated
- [ ] Docs updated


tests/e2e/settings-profile.test.ts
```diff
@@ -64,7 +64,7 @@ test('Users can update their profile photo', async ({ login, page }) => {
 	await expect(page).toHaveURL(`/settings/profile/photo`)
 
 	await page
-		.getByLabel(/change/i)
```
rajeshg, line 67: This is flaky and results in "node is not an htmlinputelement", guess file input type 
 nodes are not treated the same as text input nodes.

rajeshg:
This test assumes that the default user "kody" always exists. Is it safe to make that assumption?

kentcdodds:
Thank you!

rajeshg:
This is definitely better. Thanks for improving this further

Issue #311, opened 2023-07-15
Unable to update the profile pic when javascript is disabled


mohammedzamakhan:

kentcdodds:
Confirmed. Ideas and PR welcome :)

evayde:
The issue here is that there is no submit button, and the change event is not triggered as it runs on the client.
The most straightforward solution would probably be to add a submit button in a noscript-tag. 

Or do you know a Remix way to handle this? @kentcdodds

kentcdodds:
@evayde that sounds fine. To whoever opens the PR, I'd appreciate a video demonstrating that it works. :)

evayde:
I quickly checked it out and glimpsed over it. Unfortunately, there are some more things to consider, probably also good to keep them in mind for future challenges:

**1. UX Considerations**

This is bad:

https://github.com/epicweb-dev/epic-stack/assets/25255815/1414de63-2298-4295-b5f1-26a9f9e615de

(I have to click on edit, choose a picture, then click save). The buttons don't indicate that I have to take these steps. I know this is not really a real-world issue, as probably no user ever sees this, but still an interesting problem.

I have a possible solution for this: Since this input field is required, we can create a CSS class (but this is not possible with TW inline classes, I assume). We can use the input:valid pseudo-class to hide the submit button and only show it when the element is filled. This way we can mimic the JS behavior.
I have quickly implemented a proof of work and this is what it would look like using the CSS solution. Note that JS is disabled here.

https://github.com/epicweb-dev/epic-stack/assets/25255815/152c0248-1db2-4e2c-bda0-4ba8bc79b5cb

**2. It's possible to break the form**

The form breaks when hitting save first. Afterwards, it's not possible anymore to open the upload dialog. The user has to refresh the page:

https://github.com/epicweb-dev/epic-stack/assets/25255815/c9ef2f7f-0631-4471-a67f-da6d185cec75

This has to be further debugged; I didn't see why it's not working at first glance. But maybe it's also not a problem if we hide the save button before the input element is filled. I'd still rather have a look into it.

I think these issues should be addressed before moving forward, as an immature solution only causes us to revisit this issue at a later time.

kentcdodds:
Thanks for looking into this.

[...]

PR #321, opened 2023-07-18
fix - upload profile image without JS

evayde:
This solution enables uploading a profile image with JS disabled.
The issue is, that without JS the Save button won't show up, as it is activated using client-side JS.

The proposed solution uses the peer class on the input element and hides a server-only submit as long as the input field is invalid. `peer-invalid:[&>.server-only[type='submit']]:hidden`

The complete discussion can be found here:
https://github.com/epicweb-dev/epic-stack/issues/311

## Test Plan

- Disable JS
- Add an image using the file input
- The "Save Photo" Button shows up
- Hit "Save Photo"

## Checklist

Do we add tests for nojs implementations?

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

Before:
kodybefore

After:
kody

### DIFF
--- a/app/routes/settings+/profile.photo.tsx
+++ b/app/routes/settings+/profile.photo.tsx
@@ -182,6 +182,9 @@ export default function PhotoRoute() {
 							</label>
 						</Button>
 
+						{/* This is here for progressive enhancement. If the client doesn't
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thank you for taking the time to work on this. I think it's great.

PR #325, opened 2023-07-19
set fly concurrency to requests instead of connections

### LINKED ISSUE #317: Maybe set fly concurrency to requests instead of connections
https://github.com/epicweb-dev/epic-stack/blob/0c457b279c7839395c7b41682b4d1ca6495de1f5/fly.toml#L27

as in https://github.com/kentcdodds/kentcdodds.com/blob/d0e7ad17700361d4c052584c3143b0932c333bda/fly.toml#L33

also recommended in [fly docs](https://fly.io/docs/reference/configuration/#http_service-concurrency) for web services

CanRau:
As discussed in https://github.com/epicweb-dev/epic-stack/issues/317#issuecomment-1639241091

Hope it's ok straight against main?

closes #317

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -24,7 +24,7 @@ script_checks = [ ]
   [services.concurrency]
   hard_limit = 100
   soft_limit = 80
-  type = "connections"
+  type = "requests"
 
   [[services.ports]]
   handlers = [ "http" ]


kentcdodds (approved):
Thanks!

Issue #322, opened 2023-07-18
Improve accessibility


kentcdodds:
We have a lot of accessibility issues regarding:

1. Focus management (especially during navigations, just try to use the keyboard for a while)
2. Colors (some color contrast issues)
3. Focus outline (for sure around the buttons in the profile page and the profile image uploader)

And I expect other areas. Contributions to fix these issues is greatly appreciated.

kentcdodds:
I went through and improved a bunch of color and focus outline issues:

https://github.com/epicweb-dev/epic-stack/commit/bd9cb7482f4a1085150933004d27ca7af3cf7099
https://github.com/epicweb-dev/epic-stack/commit/1248d8465ee0fa1b6c13485b0ee5160045c1ef9c
https://github.com/epicweb-dev/epic-stack/commit/2999ed6aa6cee358a18a8300a8cf81f942e7b83d
https://github.com/epicweb-dev/epic-stack/commit/5023c871f4846900c5e45d5b3ba8e8f80024d863

PR #329, opened 2023-07-19
ui improvements - outline

rajeshg:
This is related to #322.

Minor UI improvements
1. On notes page, improve outlines to links on the left
2. Improve logout button visibility on profile page.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots
- Before
image
image
- After
image
image

### DIFF
--- a/app/routes/users+/$username.tsx
+++ b/app/routes/users+/$username.tsx
@@ -66,10 +66,7 @@ export default function UsernameRoute() {
 					{isLoggedInUser ? (
 						<Form action="/logout" method="POST" className="mt-3">
 							<Button type="submit" variant="link" size="pill">
-								<Icon
-									name="exit"
-									className="scale-125 max-md:scale-150 md:mr-1"
-								>
+								<Icon name="exit" className="scale-125 max-md:scale-150">
 									Logout
 								</Icon>
 							</Button>
--- a/app/routes/users+/$username.tsx
+++ b/app/routes/users+/$username.tsx
@@ -7,6 +7,7 @@ import { Form, Link, useLoaderData } from '@remix-run/react'
 import { GeneralErrorBoundary } from '~/components/error-boundary.tsx'
 import { Spacer } from '~/components/spacer.tsx'
 import { Button } from '~/components/ui/button.tsx'
+import { Icon } from '~/components/ui/icon.tsx'
 import { prisma } from '~/utils/db.server.ts'
 import { getUserImgSrc, invariantResponse } from '~/utils/misc.ts'
 import { useOptionalUser } from '~/utils/user.ts'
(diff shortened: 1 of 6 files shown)


app/routes/users+/$username.tsx
```diff
@@ -64,8 +65,12 @@ export default function UsernameRoute() {
...
 					{isLoggedInUser ? (
 						<Form action="/logout" method="POST" className="mt-3">
-							<Button type="submit" variant="secondary" size="pill">
-								Logout
+							<Button type="submit" variant="link" size="pill">
+								<Icon
+									name="exit"
+									className="scale-125 max-md:scale-150 md:mr-2"
+								/>
+								<span>Logout</span>
```
kentcdodds, line 73: If the Icon is intended to be rendered next to text, you should use the children feature:

```suggestion
								<Icon
									name="exit"
									className="scale-125 max-md:scale-150"
								>
									Logout
								</Icon>
```

kentcdodds (approved):
Thanks for this! Could you check that the logout icon looks ok with my suggested change? Thanks!

app/routes/users+/$username.tsx
rajeshg: Added mr-1 since, there is not much gap between icon and text.

kentcdodds: Hmmm... I prefer to keep things consistent. Could you show me a screenshot? And compare it with the delete button on a note?

rajeshg: mr-2 is what we are using currently for note-edit and note-delete.

image

Also we do not have the icon text as child. 
```jsx
<Link to="edit">
  <Icon
	  name="pencil-1"
	  className="scale-125 max-md:scale-150 md:mr-2"
  />
  <span className="max-md:hidden">Edit</span>
</Link>
```

Should we change all of them to the one below?
```jsx
<Link to="edit">
  <Icon
	  name="pencil-1"
	  className="scale-125 max-md:scale-150 md:mr-2"
  >
    <span className="max-md:hidden">Edit</span>
  </Icon>
</Link>
```

kentcdodds: I didn't realize there was a margin on the delete button. Perhaps we should update the default `gap`s on Icon 🤔 https://github.com/epicweb-dev/epic-stack/blob/84537caaec3abbbd61658aecd4ac98319eb73686/other/svg-icon-template.txt#L20-L27

other/svg-icon-template.txt
```diff
@@ -46,7 +46,7 @@ export function Icon({
 }) {
 	if (children) {
 		return (
-			<span className={`inline-flex ${childrenSizeClassName[size]}`}>
+			<span className={`inline-flex items-center ${childrenSizeClassName[size]}`}>
```
rajeshg, line 49: Adjusted gap size to 1.5 instead of the 1 currently.

Added items-center as the icon and text are not vertically centered. 

Here is the how a LARGE sized delete button looks like without items-center  
image

With items-center, it's better
image

kentcdodds (approved):
Thanks a lot!

PR #331, opened 2023-07-20
fix: export litefs-js from server file to prevent fs polyfill

kiliman:
Remix v1.19.0 now generates esbuild manifest for bundle analysis. The package `litefs-js` has a side-effect where it imports node fs module resulting in a polyfill being generated for a client bundle.

This commit exports litefs-js from a *.server file to ensure that the polyfill is not bundled.

### Bundle size 2.1 mb, 266 kb from polyfill in routes/admin+cache
image

image

### Bundle size 1.8 mb (no more polyfill)
image

### DIFF
--- a/app/utils/litefs.server.ts
+++ b/app/utils/litefs.server.ts
@@ -1,0 +1,2 @@
+export * from 'litefs-js'
+export * from 'litefs-js/remix.js'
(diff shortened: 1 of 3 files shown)


app/utils/litefs.server.ts
```diff
@@ -0,0 +1,2 @@
+export * from 'litefs-js'
```
kentcdodds, line 1:
```suggestion
// litefs-js should be used server-side only. It imports `fs` which results in Remix
// including a big polyfill. So we put the import in a `.server.ts` file to avoid that
// polyfill from being included. https://github.com/epicweb-dev/epic-stack/pull/331
export * from 'litefs-js'
```

kentcdodds (approved):
Thank you!

PR #332, opened 2023-07-21
dev: add sly

jacobparis:
## Test Plan

There are new instructions in the docs to add icons using Sly. Run the command there and you should see new icons appear in `./other/svg-icons` and the `<Icon>` component's types will update automagically.

- I have put the `sly.json` file inside `/other/sly`. I can move it elsewhere if need
- `/other/sly` also contains the transformer for the icons. I figured out how to make the transformers typescript without requiring any weird flags in the command so that works out of the box now.
- The transformer currently removes the width/height attributes from the SVG, which is technically not needed for now while we're using rmx-cli to generate the sprite. I'm exploring options to bring more of that functionality within the scope of Sly
- The transformer also prepends the github URL and license for each icon. Radix is MIT which doesn't require a copyright notice on larger works (which Epic Stack should qualify as). Lucide Icons on the other hand is ISC and I think that means it DOES require the license. For that reason I've included this as part of the transformer example and folks are free to change as they want. I do think there's value in showing which repo an SVG came from to help find similar icons later, in case they start to get mixed up.

## Checklist

- [X] Docs updated


kentcdodds (changes requested):
This is awesome!

Two things:

1. Since we're already removing the width/height with rmx-cli, I think we can drop the node-html-parser dep and not do that ourselves.
2. Do you think you could update all the icons we have already so they have the license info?

If not, I can do that later.

jacobparis:
On it

jacobparis:
All fixed – just took the one line 

```
npx sly add @radix-ui/icons arrow-left arrow-right avatar camera check clock cross-1 exit file-text laptop lock-closed lock-open-1 magnifying-glass moon pencil-1 pencil-2 plus reset sun trash update --overwrite --yes
```

kentcdodds (approved):
I love this. Thank you!

docs/icons.md
```diff
@@ -6,6 +6,24 @@ You'll find raw SVGs in the `./other/svg-icons` directory. These are then
...
+```
+
+If you don't specify the icons, Sly will show an interactive list of all the
+icons available in the `@radix-ui/icons` collection and let you select the ones
+you want to add.
+
+Sly has been configured in the Epic Stack to automatically add the icons to the
+`./other/svg-icons` directory and to run the `npm run build:icons` script
+automatically, so there are no extra steps to take. You can see the
+configuration in the `./other/sly/sly.json` file.
```
kentcdodds, line 25: Oh, just occurred to me... How does sly find this config file? Does it just look for `sly.json` anywhere in the project? Because that would be brilliant.

jacobparis: https://github.com/jacobparis-insiders/sly/blob/main/cli/src/get-config.ts#L14

I check a few different directories and patterns, open to adding more for sure 

I should add a --config flag and environment variable that lets you specify a specific path though

kentcdodds: This is great 👍 Thank you for starting out the project with the ability to put config somewhere other than the root directory 🙏

PR #334, opened 2023-07-22
fix: bring build-icons into the epic stack

jacobparis:
## Test Plan

I've created a `./build-icons.ts` file that generates the svg spritesheet

I have **deleted the icon component template** and updated the gitignore to commit the real Icon.tsx. In this PR it shows that as a rename, which would be easy to overlook.

We don't need to gitignore it anymore because **the Icon.tsx is no longer being codegen'd**. Instead, we generate a JSON file with the names of the icons, and import it as a type to get all the icon names as a string union. This new JSON file is gitignored.

image

## Things to note

- This replaces the implementation of npm run build:icons script, but everyone can continue calling that script normally
- Sly doesn't require any reconfiguration to keep working this way either
- This removes the rmx-cli dev dependency
- This adds a `node-html-parser` dev dependency. I prefer this approach for robustness and flexibility, but it would be possible to replace it with a bunch of regex matchers to transform the SVG as a string (into symbols for the spritesheet), if you prefer.
- The JSON file looks like this. I used an object for the iconNames because I think that's the only way to get types off the keys. An array would work if we could mark it `as const`, but that's not possible when importing a JSON file. 
```json
{
	"_comment": "This file is generated by npm run build:icons",
	"iconNames": {
		"update": true,
		"trash": true,
		"sun": true,
		"reset": true,
		"plus": true,
		"pencil-2": true,
		"pencil-1": true,
		"moon": true,
		"magnifying-glass": true,
		"lock-open-1": true,
		"lock-closed": true,
		"laptop": true,
		"file-text": true,
		"exit": true,
		"cross-1": true,
		"clock": true,
		"check": true,
		"camera": true,
		"avatar": true,
		"arrow-right": true,
		"arrow-left": true
	}
}
```

## Checklist

- [ ] Tests updated
- [ ] Docs updated


other/build-icons.ts
```diff
@@ -0,0 +1,125 @@
...
+
+	const output = [
+		`<?xml version="1.0" encoding="UTF-8"?>`,
+		`<!-- This file is generated by npm run build:icons -->`,
+		`<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="0" height="0">`,
+		...symbols,
+		`</svg>`,
+	].join('\n')
+
+	const existingOutput = await fs.readFile(outputPath, 'utf8')
```
jacobparis, line 91: This introduces an extra read every time in order to save us from writing sometimes. I'm not actually sure this is a net benefit. Maybe to avoid locking if there were lots of processes trying to write this file, but that's not likely to be the case here

other/build-icons.ts
```diff
@@ -0,0 +1,125 @@
...
+		_comment: `This file is generated by npm run build:icons`,
+		iconNames: Object.fromEntries(
+			files.map(file => {
+				// We can only get type inference on the keys of a json file
+				return [path.basename(file, '.svg'), true]
+			}),
+		),
+	}
+
+	const json = JSON.stringify(output, null, '\t')
```
jacobparis, line 122: this is my first time ever writing a stringify with a tab character

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "postinstall": "patch-package --patch-dir ./other/patches",
     "build": "run-s build:*",
-    "build:icons": "rmx svg-sprite other/svg-icons app/components/ui/icon.tsx --template=./other/svg-icon-template.txt",
+    "build:icons": "tsx ./other/build-icons.ts",
```
jacobparis, line 12: I wasn't quite sure where to put this file. In the svg-icons folder made sense, in the Sly folder kinda made sense, but this spot matches the build:server command so I copied the filename format of that too

app/components/ui/icon.tsx
```diff
@@ -1,10 +1,10 @@
-// THIS FILE IS GENERATED, edit ./other/svg-icon-template.txt instead
-// then run "npm run build:icons"
-
 import { type SVGProps } from 'react'
 import { cn } from '~/utils/misc.ts'
 import href from './icon.svg'
+
+import type { iconNames } from './icons.json'
 export { href }
+export type IconName = keyof typeof iconNames
```
kentcdodds, line 7: Ok, here me out... What if we make this type somehow resilient to whether there is an `icons.json` file built yet and if it's not it just defaults to `any` or something that will make TypeScript not fail the build if we've not yet generated the icons.json? 🤔

kentcdodds (changes requested):
@kiliman actually added a new feature to rmx-cli that gets us pretty close to what you've got here, but I do really like the flexibility this offers folks so I think I want to pursue this (sorry @kiliman!).

I just have one thought.

kiliman:
No offense but if you wanted to inline this code you could have asked me. This is now the second time that a feature I wrote (first being time zone client-hint) was added by someone else.

jacobparis:
I'm sorry @kiliman, I didn't mean to step on any toes. I started looking into this so I could have a general solution for rendering icons with Sly for [my own projects](https://www.jacobparis.com/content/svg-icons), but I didn't want to codegen the react bits. The epic stack was just a testbed for following my own instructions and making sure they work, and it did so I proposed the PR.

I'm sorry if it felt I was undermining your contributions here, and I should have reached out before making the PR.

app/components/ui/icon.tsx
jacobparis: The easiest way (only way?) to do that is to add it to the types/deps.d.ts as a module. Typescript will pick it up if it can't find 

https://github.com/epicweb-dev/epic-stack/assets/5633704/10bb9ec7-e600-4c73-b14a-afb662869ab5

Do you think deps.d.ts makes sense or should it be its own definition file?
```ts
declare module '*/icons.json' {
	// This is a fallback for when the /app/components/ui/icons.json file is not generated
	const iconNames: Record<string, string>
	export { iconNames }
}
```

kentcdodds: Yeah, I think that's alright 👍 good idea!

kentcdodds:
> This is now the second time that a feature I wrote (first being time zone client-hint) was added by someone else.

@kiliman this comes across as if you're bitter. I'm sorry if that's the case. I understand if this makes you less likely to want to contribute the the Epic Stack in the future.

I will always go the direction that I think is better, [even with my own stuff](https://kentcdodds.com/blog/favor-progress-over-pride-in-open-source). In this case, I'm going to proceed with Jacob's approach.

app/components/ui/icon.tsx
jacobparis: Lots of folks on twitter have been asking why not just generate the type, and the reason is to avoid needing more files with disclaimers not to touch them

But if we generate it to node_modules (like prisma does with .client) or build or somewhere already designated as an output directory, a ts file makes a lot of sense to me

How do you feel about that approach instead?

kentcdodds: Yes. I definitely prefer that. Would reduce the likelihood of someone accidentally editing the file directly

jacobparis: Ok so it looks like node_modules is off limits because any type definition file (required to keep the import from breaking if the file doesn't exist) will override node_modules even if it does exist

Anywhere outside of that works. Top candidates are
- `/build/icon-names.ts`
- `/types/.icon-names.ts`

However, there's no clean way to import from that directory. Our only path aliases are ~ for the app directory and one for the tests directory

I don't love this sort of semi-relative path if it can be avoided

```ts
import type { IconName } from '~/../types/.icon-names.ts'
```

Options are, in my order of preference,
- build directory, accept the weird import
- types directory, accept the weird import
- either of the above with a new path alias for the root directory `@/build/icon-names.ts`
- hidden file in the root of app? `~/.icon-names.ts`

rajeshg:
> 

@kiliman my apologies. Didn't mean to step on your toes with the timezone client hints feature.

kiliman:
Hey everyone, I'm sorry if I made you all feel bad. I shouldn't have posted what I did. It was a flash of annoyance, and I should have just moved on.

Anyway, I hope we're all good. The ultimate goal is to make this the MOST epic stack, so how the code gets there should not be a concern.

Again, my apologies. Thanks for your understanding.

kentcdodds:
I'm glad you're here @kiliman. Thanks for the good work you do!

kentcdodds:
I opened a new PR based on this one (because I can't push to @jacobparis's `main` branch). I figured out how to make the type fallback work with `paths`. It's pretty slick. So TypeScript won't cry about the Icons not being generated right from the get-go, but we generate them as part of the build.

We still probably better check whether the icons have been generated (and generate them if they haven't) as a pre-dev script. But I'm happier with this.

kentcdodds:
This was merged in another PR.

PR #339, opened 2023-07-24
fix: issue where user could not edit the password and user name at same time. close #335

### LINKED ISSUE #335: Confirming current password is using `username` provided by the user
There's a bug on the setting new password action. It's using the `username` provided by the user to check if the current password user is valid but if you set a new username at same time as you change the password you get an invalid password error

https://github.com/epicweb-dev/epic-stack/blob/main/app/routes/settings%2B/profile.index.tsx#L73C6-L73C6

wladiston:
We were checking the user's username based on the user input to validade their current password. This PR changes it to check the user's current password based on the userId.

## Test Plan

Go tot the settings page and try to change your username and password at same time

## Checklist

- [x] Tests updated

### DIFF
--- a/app/routes/settings+/profile.index.tsx
+++ b/app/routes/settings+/profile.index.tsx
@@ -11,7 +11,7 @@ import {
 	authenticator,
 	getPasswordHash,
 	requireUserId,
-	verifyLogin,
+	verifyUserPassword,
 } from '~/utils/auth.server.ts'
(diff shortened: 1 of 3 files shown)


kentcdodds (approved):
Even better than what I was thinking. Looks great. Thanks!

wladiston:
Had a good teacher

PR #340, opened 2023-07-25
fix: bring build-icons into the epic stack

kentcdodds:
This is a new PR because I couldn't update #334 because it was made from @jacobparis's `main` branch on his fork and I can't push to other people's main branch.


jacobparis:
> This is a new PR because I couldn't update #334 because it was made from @jacobparis's `main` branch on his fork and I can't push to other people's main branch.

I had no idea this was a constraint until I saw your tweet about it yesterday, I'll use real branches from here on out

```ts
    "@/icon-name": [
      "app/components/ui/icons/name.d.ts",
      "types/icon-name.d.ts"
    ]
```

I didn't know this was possible! I would not have come up with this solution on my own 

It does raise one alternative of committing the name.d.ts hardcoded to string and then gitignoring any changes while the build:icons overwrites it. That'd keep it all in a single file, though I don't know if future git pulls would still try to pull the ignored file over local changes?

Either way I'm happy with where this PR has ended up

kentcdodds:
Thanks a lot for your help!

PR #341, opened 2023-07-25
chore: update remix packages

Klukies:
I've update the remix packages and, as the [release notes](https://github.com/remix-run/remix/releases/tag/remix%401.19.0) state, have updated `useRouteLoaderData` to make use of the generic type within the application.


kentcdodds (approved):
Nice! I missed that bit in the release notes. Thanks!

app/utils/request-info.ts
```diff
@@ -1,11 +1,10 @@
-import { type SerializeFrom } from '@remix-run/node'
 import { useRouteLoaderData } from '@remix-run/react'
 import { type loader as rootLoader } from '~/root.tsx'
 
 /**
  * @returns the request info from the root loader
  */
 export function useRequestInfo() {
-	const data = useRouteLoaderData('root') as SerializeFrom<typeof rootLoader>
+	const data = useRouteLoaderData<typeof rootLoader>('root')!
```
kentcdodds, line 8: I strongly prefer avoiding `!`. Could we use our `invariant` util instead?

Klukies: Aha, I've been working on a project that currently isn't using `invariant` so didn't jump to it.
I've added now 😊

kentcdodds (approved):
Thank you!

PR #342, opened 2023-07-25
use zod to parse environment variables

stephen776:
Use zod to create and parse a schema that represents the required server environment variables. The main benefit here is that the  schema can explicitly declare and validate complex types for env variables and potentially coerce strings into the appropriate types where needed(such as numbers and enum values.)

### DIFF
--- a/app/utils/env.server.ts
+++ b/app/utils/env.server.ts
@@ -1,28 +1,34 @@
-import { invariant } from './misc.ts'
+import { z } from 'zod'
 
-const requiredServerEnvs = [
-	'NODE_ENV',
-	'DATABASE_PATH',
-	'DATABASE_URL',
-	'SESSION_SECRET',
-	'INTERNAL_COMMAND_TOKEN',
-	'CACHE_DATABASE_PATH',
-	// If you plan to use Resend, uncomment this line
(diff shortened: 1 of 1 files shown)


app/utils/env.server.ts
```diff
@@ -1,28 +1,34 @@
...
-	'DATABASE_URL',
-	'SESSION_SECRET',
-	'INTERNAL_COMMAND_TOKEN',
-	'CACHE_DATABASE_PATH',
-	// If you plan to use Resend, uncomment this line
-	// 'RESEND_API_KEY',
+import { z } from 'zod'
+
+const schema = z.object({
+	NODE_ENV: z.enum(['production', 'development', 'test'] as const),
```
kentcdodds, line 4: I'm concerned about the `enum` use here. Does that mean I can't do this:

```
if (process.env.NODE_ENV === 'development') {
}
```

With enums you normally have to reference some type for the equality check. Normally I would do this instead:

```suggestion
	NODE_ENV: z.union([z.literal('production'), z.literal('development'), z.literal('test')]),
```

kentcdodds (commented):
Thanks for this. Just one thing.

PR #343, opened 2023-07-25
Fix css links issues due to RDT

xHomu:
RDT css was accidentally leaked into the production build, making it a render blocking resource and delaying load time:

image

https://www.webpagetest.org/video/compare.php?tests=230725_AiDcJA_B64-r:2-c:0

## Test Plan

As this is a browser behavior, I don't believe there is a test for this.

## Checklist

- [x] Tests updated
- [x] Docs updated

### Before
image

### After
image


kentcdodds (approved):
Thank you!

PR #344, opened 2023-07-26
Remix Development Tools hmr issue

AlemTuzlak:
Fixes issue with HMR for remix development tools

Note:
This is a quick and dirty solution, I will still be investigating the issue and fixing it on the repo then creating a PR to fix it here as well!

## Test Plan

do a hmr => works

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

Issue #349, opened 2023-07-27
build-icons script does not always detect new icons


DigitalGoldfish:
This is the code the script uses to determine if the icon files should be updated:
```
const spriteUpToDate = iconNames.every(name =>
		currentSprite.includes(name),
	);
const typesUpToDate = iconNames.every(name => currentTypes.includes(name));
```

This has multiple issues:
 - If a  new icon is a substring of an already existing icon it doesn't get added (e.g. if double-arrow-left already exist arrow-left doesn't get added)
 - If an icon is modified the modified svg will not be copied into the sprite.svg
 - Removed icons are not removed from the sprite.svg/name.d.ts

The first issue could be fixed by modifying the checks like this 
```
const spriteUpToDate = iconNames.every(name =>
		currentSprite.includes(`id="${name}"`),
	);
const typesUpToDate = iconNames.every(name => currentTypes.includes(`"${name}"`));
```

kiliman:
In `rmx-cli`, I compare the "generated" file with the existing file, if present, and will not overwrite it. This ensures that any changes will update the file, but will not trigger file watchers with false positives.

kentcdodds:
This is fixed.

PR #350, opened 2023-07-27
Improve detection of new svg icons #349

DigitalGoldfish:
Improve the check if an svg icon is already contained in the spritesvg/name.d.ts by making sure there is no icon with the same name but a pre- or suffix. 

## Test Plan

-

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (commented):
Good call 👍 Thanks!

PR #351, opened 2023-07-27
fix: exclude doctypes + xml headers from svg sprite

jacobparis:
There was a bug report from @AlemTuzlak today that if there's an SVG that has its own doctype or xml header, it breaks the generated spritesheet.

This is fixed by only returning the svg element (which has been transformed to symbol) instead of the root of the syntax tree during the `generateSvgSprite` function. Without this change, the doctypes and xml headers for every symbol are copied into the sprite producing an invalid output. This change is in the first commit

Also adds a small change to append a trailing newline to the SVG file. This is [for POSIX compliance](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) and helps keep git diffs clean as many devs have systems set up to automatically add these

## Test Plan

The second commit contains added tests to ensure this behaviour doesn't regress

I have added two tests: one that snapshots two icons turning into the expected spritesheet, and an additional one that covers this case specifically ensuring no doctypes or additional xml headers have made it into the spritesheet.

I'm open to feedback on the structure of the test. Possible changes that come to mind
- mocking the file system instead of reading and writing files
- move setup and cleanup into each test instead of beforeAll and afterEach. I think we need to use the hooks to ensure the files are cleaned up after a failed test
- testing each aspect of the svg generation, like missing width/height attributes, IDs being present, and being named symbol, instead of using an inline snapshot
- moving the test file to /tests instead of amending the vitest config to also read tests in /others

## Checklist

- [x] Tests updated


kentcdodds (changes requested):
Thanks for the fix! I'm not so certain we need to worry about tests for this. It's really unlikely to be changed and it's a relatively simple script 🤷‍♂️

jacobparis:
Ok perfect, I'll remove the tests shortly

kentcdodds (approved):
Thanks!

PR #353, opened 2023-07-28
Updated Remix Dev Tools to latest version

AlemTuzlak:
Latest version of Remix Dev Tools
- Important update due to stabilization of showing route boundaries (it was causing re-renders before and was generally less stable)
## Test Plan

Run app => see route boundaries, no re-renders, no submissions, no hydration errors

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Cool 👍 Thanks!

Issue #354, opened 2023-07-30
UI glitch when opening user profile


kandros:
When clicking on a user from the list at http://localhost:3000/users, I've noticed a glitch that makes the page scroll to the top before the navigation.

~~I'm not yet familiar enough with remix and the codebase to fix it myself at the moment, but I figured out it's caused by the autofocus of in search bar input https://github.com/epicweb-dev/epic-stack/blob/main/app/components/search-bar.tsx#L46~~ opened PR

Here's a video with slow motion frames that shows it
https://github.com/epicweb-dev/epic-stack/assets/4562878/6be1e0cb-57de-4c42-b773-821e2c8c5fa5

lewisblackburn:
This happens when opening dialogues and dropdown menus too

kentcdodds:
I've removed the dev tools temporarily until this gets fixed.

PR #355, opened 2023-07-30
fix: ui glitch when navigatin to user detail (#354)

kandros:

### DIFF
--- a/app/routes/users+/index.tsx
+++ b/app/routes/users+/index.tsx
@@ -4,7 +4,7 @@ import {
 	type DataFunctionArgs,
 	type HeadersFunction,
 } from '@remix-run/node'
-import { Link, useLoaderData } from '@remix-run/react'
+import { Link, useLoaderData, useNavigation } from '@remix-run/react'
 import { z } from 'zod'
 import { GeneralErrorBoundary } from '~/components/error-boundary.tsx'
 import { SearchBar } from '~/components/search-bar.tsx'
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Nice catch. Thanks!

kentcdodds:
Hmmm.... I want to try a few things with this to make sure there aren't intended consequences.

kentcdodds:
Yeah, this isn't the fix for the issue. The issue is that the SearchBar is getting unmounted and re-mounted. I'm not sure what's causing it just yet, but if we figure that out this issue will probably go away. Either way, we need to figure that out first before we merge this.

kentcdodds:
This is caused by [Remix Dev Tools](https://github.com/Code-Forge-Net/Remix-Dev-Tools). Removing it resolves the bug. @AlemTuzlak, could you please figure out why routes are getting unmounted and remounted on search param changes?

AlemTuzlak:
@kentcdodds will do! I think this will probably be mitigated by the new way of initializing the dev tools I'm working on. Will investigate it further as well

AlemTuzlak:
@kentcdodds I've been taking a very deep dive into Remix internals the last couple of days and have figured out some issues with my approach to rendering the component, I am working on a release that will land soon that will not cause re-renders or issues on the UI at all because the component will be rendered in a completely different way instead of it being in the root. I use the epic stack as a testing env in remix-development-tools and I can confirm this issue is gone with the new approach. Will be making a PR to address this asap

kentcdodds:
I've removed the dev tools temporarily until this gets fixed.

AlemTuzlak:
@kentcdodds FYI I found the root cause of this, will be fixed in the upcoming v2 along with some more stability improvements!

PR #359, opened 2023-08-01
fix tests when using resend

iainfreestone:
When setting up resend for emails the tests fail after changing the email address in `app/utils/email.server.ts`

## Checklist

- [ ] Tests updated
- [x] Docs updated


kentcdodds (commented):
👍👍 Thanks!

PR #362, opened 2023-08-02
move faker to prod deps so that it can be used during prod seed

sys13:


kentcdodds:
This is not necessary. I still need to take the time to write the docs in Epic Stack, but here's how I'm recommending seeding in production: https://github.com/epicweb-dev/data-modeling/tree/main/exercises/04.seed

We don't want to install faker in prod, so I'm going to close this. Thanks!

PR #366, opened 2023-08-03
add banner ui components

vinstah:
Adding a global banner to root.tsx and other banner components in ui folder. The banner contents can be managed in a CMS in the future.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

image


kentcdodds:
Thanks for working on this! I don't want to have this built-in, but I'd love an example if you'd like to make one for the docs

PR #372, opened 2023-08-04
Remix Development Tools 2.0

AlemTuzlak:
Added back V2 of Remix Development Tools into the project!

## Test Plan

Try changing the theme, and also navigating to the search page =>there should be no flickers
- No console errors
 - settings page should also be there


AlemTuzlak:
@kentcdodds Hey Kent! Don't want to be intrusive but just wanted to check on the status on this and if you need any clarifications or anything else! You might just be super busy too.

kentcdodds:
Hi @AlemTuzlak,

I want to try this out, but I need it to be updated to the latest. Could you please update this PR to be a single commit on top of the latest `main`? The easiest way to do this would be to create a new branch.

In the future, I recommend you make a new branch for every PR rather than merging `main` into an existing branch. It makes tracking changes and managing git *much* easier.

Whenever you see stuff like this:

image


You know that it's going to be hard to manage updates in `main`.

AlemTuzlak:
@kentcdodds will do, sorry I'm still learning all this forking stuff, never did it before, mind trying it out? It will be the same stuff when I branch off of main, will try to do it tonight

PR #373, opened 2023-08-04
Clarify location of SENTRY_AUTH_TOKEN in Sentry UI

skjelmo:
Minimize Setup Friction by clarifying which token to select as the `SENTRY_AUTH_TOKEN`.

## Test Plan

N/A

## Checklist

- [ ] Tests updated
- [x] Docs updated

## Description
Prevent ambiguity between the `SENTRY_AUTH_TOKEN` at the bottom of the page and the `Client Secret` located at the very bottom of the page which looks like a token without being the correct one. 

I spent a while debugging why the deployment failed before realizing that I had selected the wrong token from Sentry. Having fly.io managing secrets in the Github Action was new to me, so I wasted time checking if it had to be added as Github secrets before realizing my mistake. I hope this small clarification will help prevent future users from making the same mistake.  

## Screenshots
**From Sentry:**
Screenshot 2023-08-04 at 19 37 56

### DIFF
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -26,7 +26,7 @@ Once you see the onboarding page which has the DSN, copy that somewhere (this
 becomes `SENTRY_DSN`). Then click
 [this](https://sentry.io/orgredirect/settings/:orgslug/developer-settings/new-internal/)
 to create an internal integration. Give it a name and add the scope for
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #374, opened 2023-08-05
move hsl values to tailwind.config for editor color previews

sys13:
image

It's nice to have a color preview and picker in editors that support it


EmilEinarsen:
The CSS variables need to be defined as channels with no color space function to work with the [opacity modifier syntax](https://tailwindcss.com/docs/text-color#changing-the-opacity).
See [Customizing Colors](https://tailwindcss.com/docs/customizing-colors#using-css-variables)

sys13:
Thank you @EmilEinarsen for the insight!

PR #382, opened 2023-08-10
Upgrade conform version

edmundhung:
Upgraded conform to [v0.8](https://github.com/edmundhung/conform/releases/tag/v0.8.0) with schema adjusted for the new automatic type coercion feature :fire:

## Test Plan

I have test the following forms manually:
- Login (with 2fa)
- Forgot password / Reset password
- Onboarding
- Profile update
- Enable 2fa
- Update profile picture
- Create / Update Note

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

n/a


app/routes/_auth+/onboarding.tsx
```diff
@@ -38,24 +37,30 @@ const OnboardingFormSchema = z
...
-	})
+			.superRefine(({ confirmPassword, password }, ctx) => {
+				if (confirmPassword !== password) {
+					ctx.addIssue({
+						path: ['confirmPassword'],
+						code: 'custom',
+						message: 'The passwords did not match',
+					})
+				}
+			}),
```
edmundhung, line 62: Zod does not run `refine` or `superRefine` until all fields are provided. This makes it run earlier as soon as both `password` and `confirmPassoword` is filled.

Ref: https://github.com/colinhacks/zod/issues/479

app/routes/_auth+/forgot-password/index.tsx
```diff
@@ -90,7 +89,7 @@ export async function action({ request }: DataFunctionArgs) {
 	if (response.status === 'success') {
 		return redirect(redirectTo.toString())
 	} else {
-		submission.error[''] = response.error.message
+		submission.error[''] = [response.error.message]
```
edmundhung, line 92: The `report` helper is not available yet. But the error has to always be an array now.

kentcdodds: Sounds good 👍

app/routes/_auth+/onboarding.tsx
kentcdodds: Oh fantastic. Great idea. Thank you!

app/routes/resources+/note-editor.tsx
```diff
@@ -14,16 +14,15 @@ import { redirectWithToast } from '~/utils/flash-session.server.ts'
 
 export const NoteEditorSchema = z.object({
 	id: z.string().optional(),
-	title: z.string().min(1),
-	content: z.string().min(1),
+	title: z.string(),
+	content: z.string(),
```
kentcdodds, line 18: I seem to remember talking about this, but I forgot. Does this mean that users will be able to create notes with no title or content? Or is `min(1)` assumed? If it's assumed, then I would need to do `z.string().optional()` if I wanted to allow an empty string? I think that makes sense in this context, but I just wanted to make sure my understanding is correct.

kentcdodds (approved):
Thank you! I'm so grateful to you for taking the time to contribute this upgrade!

kentcdodds:
Will have to check in on why playwright's not working. I'll kick it off again to see if it was a fluke (😬).

app/routes/resources+/note-editor.tsx
edmundhung: It's adhere to how zod works now which means all fields are required by default. 

>  I would need to do z.string().optional() if I wanted to allow an empty string
Exactly

edmundhung:
Happy to see the zod schemas getting simplified with this! Love it. 😄 I pull the fix you made on main branch and the tests should be good now.

PR #383, opened 2023-08-11
feat: add github actions extensions

xstevenyung:
adding 2 additionals recommanded extensions.
- `github.vscode-github-actions` → official github actions extensions, really useful while working on customizing the workflows
- `tamasfe.even-better-toml` → useful for the `fly.toml` file (not sure if this one is really necessary as there is only the `fly.toml` that benefits from it)

## Test Plan

not needed

## Checklist

- [X] Tests updated
- [X] Docs updated

## Screenshots

not needed

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -5,6 +5,7 @@
     "esbenp.prettier-vscode",
     "prisma.prisma",
     "qwtel.sqlite-viewer",
-    "yoavbls.pretty-ts-errors"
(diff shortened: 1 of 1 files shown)


kentcdodds:
Love the actions one. Let's drop the toml one though since people will rarely need it.

Thanks!

xstevenyung:
updated ✅

kentcdodds (approved):
Thanks!

Issue #391, opened 2023-08-17
Bug: development build not reloading after changes


mikesurowiec:
The problem appears to be here: https://github.com/epicweb-dev/epic-stack/blob/main/server/index.ts#L37

Changing it back to the relative path allows it to work again `const BUILD_PATH = '../build/index.js'` ✨

PR #392, opened 2023-08-17
fix: development build reloading

mikesurowiec:
Closes: https://github.com/epicweb-dev/epic-stack/issues/391

The alias import doesn't work here, changing it back to a relative import fixes the development build reload.

## Test Plan

I manually confirmed that making a change resulted in the page live-reloading in development mode.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Sounds good. Thanks!

PR #393, opened 2023-08-18
Update misc.tsx

SomiDivian:
update `useIsPending` hook to be conditionally enabled

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds:
Thanks! Can you show an example of why this would be necessary?

SomiDivian:
https://github.com/epicweb-dev/epic-stack/pull/394#issuecomment-1684986106

kentcdodds:
Sorry, looks like this won't be happening. Thanks for the PR anyway!

PR #403, opened 2023-08-25
add example with fanthom analytics

xstevenyung:
simple fanthom analytics example via CDN, would love feedback if there is stuff missing on this. i'm currently implementing this on my project so any improvement would be greatly appreciated

## Checklist

- [X] Tests updated
- [X] Docs updated

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -53,7 +53,10 @@ This page links to examples of how to implement some things with the Epic Stack.
   by [@kentcdodds](https://github.com/kentcdodds): An example of the Epic Stack
   with OpenID Connect Auth (Google) using [web-oidc](https://npm.im/web-oidc)
   and [remix-auth](https://npm.im/remix-auth).
-
+- [Epic Stack + Fathom Analytics](https://github.com/xstevenyung/epic-stack-with-fathom-analytics)
+  by [@xstevenyung](https://github.com/xstevenyung): An example of the Epic Stack
+  with Fanthom Analytics via CDN
+  
 ## How to contribute
 
 [![Kent screencast showing the examples page](https://github.com/epicweb-dev/epic-stack/assets/1500684/7074f1db-c918-42c6-a724-0b082168395f)](https://www.epicweb.dev/tips/contribute-an-epic-stack-example)


kentcdodds (commented):
Thanks for this! There are a few more things that should probably be handled in your example:

1. Update the CSP to support the fathom script: https://github.com/kentcdodds/kentcdodds.com/blob/5dc7a5e6f0561023be3b348c5bf5fd88f3391eb3/server/index.ts#L247
2. Handle client-side pageviews: https://github.com/kentcdodds/kentcdodds.com/blob/5dc7a5e6f0561023be3b348c5bf5fd88f3391eb3/app/root.tsx#L384-L419
3. Handle pre-fathom-loaded function calls with a fathom queue: https://github.com/kentcdodds/kentcdodds.com/blob/5dc7a5e6f0561023be3b348c5bf5fd88f3391eb3/app/root.tsx#L525-L535

Hope that helps! Thanks for making the example!

xstevenyung:
made some changes to implement everything but made some modification and would love your feedback again to suit the epic stack to the maximum:

- move all the code (script tag component + code related to fathom) into a `utils/fathom.tsx`
- i saw that your `<script />` for fathom on your repo uses `onLoad`, from all my testing the `onLoad` never triggered. i made some research on this but couldn't confirm if this is a bug or if i screw up on my end. for the time being, i moved everything into a load event listener until i can figure it out
- also added all the other methods available in fathom (`trackGoal`, `enableTrackingForMe`, etc.)
- change the queue system from command (which was hard to type) to queue of function as we are on the client-side only. this ease the typescript and execution of the queue on load.

kentcdodds:
Looking great. I love the improvements! I left a couple comments.

xstevenyung:
made it work with a proxy and also changed the `useFathom` to return a getter function to be able to get the fathom instance or the queue proxy at the moment of using it.

not sure how I feel about it when I feel like this is fine. it's a lot of complexity added for little benefits but I will let you be the judge.

if there is any other way you have in mind, would take it.

I made a separate PR to easily compare the two options
https://github.com/xstevenyung/epic-stack-with-fathom-analytics/pull/1/files

kentcdodds (approved):
Love it! Once you merge that proxy solution then I'll merge this 👍👍 Thanks a lot! I think we'll probably bring this into the core eventually.

xstevenyung:
updated 👌

kentcdodds:
Thanks!

PR #404, opened 2023-08-26
Clean up inserted user on login e2e test

erikras:
Perhaps I'm missing something, but I was reading through the e2e test code, and in `onboarding with link`, we're clearly remembering the username we've signed up with to delete it after the test is over, but in `login as existing user`, we're inserting a user without marking it for later deletion. Maybe this isn't a problem because the e2e tests always start with a freshly seeded db? Or is this an oversight?

## Checklist

- [X] Tests updated
- [ ] Docs updated

### DIFF
--- a/tests/e2e/onboarding.test.ts
+++ b/tests/e2e/onboarding.test.ts
@@ -127,6 +127,7 @@ test('onboarding with a short code', async ({ page }) => {
 test('login as existing user', async ({ page }) => {
 	const password = faker.internet.password()
 	const user = await insertNewUser({ password })
+	usernamesToDelete.add(user.username)
 	invariant(user.name, 'User name not found')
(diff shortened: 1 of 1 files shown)


tests/e2e/onboarding.test.ts
```diff
@@ -127,6 +127,7 @@ test('onboarding with a short code', async ({ page }) => {
 test('login as existing user', async ({ page }) => {
 	const password = faker.internet.password()
 	const user = await insertNewUser({ password })
```
kentcdodds, line 129: This utility handles cleanup for us 😁

But I have noticed sometimes we're not cleaning up the users properly. So if you can figure out why the cleanup logic doesn't always work then I'll gladly merge a PR! But this one isn't the proper fix. Thanks though!

PR #407, opened 2023-08-28
Standardizing Environment Checks: MODE vs. NODE_ENV

MoSattler:
While looking through the code, I saw that some files use `MODE` and others use `NODE_ENV` for checking the environment. Specifically, in this entry.client.ts file, we seem to be using both (see [this](https://github.com/epicweb-dev/epic-stack/pull/407/files#diff-c5698d7b301af51f1573ea4b61bbd1878710472917906bf4f779a64fc3055615R5-R7)). I've tried sticking to just `MODE` to keep it simple. Is there a reason we use both ways? And would it be okay to use just `MODE` in most places?

### DIFF
--- a/app/entry.client.tsx
+++ b/app/entry.client.tsx
@@ -11,7 +11,7 @@ const callback = () =>
 		hydrateRoot(document, <RemixBrowser />)
 	})
 
-if (process.env.NODE_ENV === 'development') {
+if (ENV.MODE === 'development') {
 	import('remix-development-tools').then(({ initClient }) => {
 		// Add all the dev tools props here into the client
 		initClient()


kentcdodds (approved):
This is definitely what we should be doing. Thanks!

kiliman:
`esbuild` will handle `process.env.NODE_ENV` at **build-time**. So, in this instance, the code to import `remix-development-tools` will only be included if `NODE_ENV === 'development'`. For production, it will be `false`, and the code will be removed from the bundle.

`ENV.MODE` is a runtime check, so the code will always be bundled.

kentcdodds:
It's a dynamic import so it'll be in a separate chunk, but yeah that's right. We should put a comment in place instead.

I'm pretty close to ripping out the development tools. It bothers me how invasive they are in the codebase and in the app as a whole... I wish it worked more like the React dev tools and could just be an extension that required no app-level integration.

kiliman:
@kentcdodds true, but `esbuild` will remove the entire `if` block, if the expression is false, so less code in the bundle.

As for dev tools, I understand the frustration, but it's kind of a limitation of Remix/SSR/Hydration. I think if you don't care about rendering route boundaries, you probably don't need the boilerplate.

PR #408, opened 2023-08-28
fix notes image upload

arpitdalal:
Image upload for notes has a bug where _only_ adding an image to an existing note **deletes** the old one and _only_ adds the new one. (See below video)

https://github.com/epicweb-dev/epic-stack/assets/61059807/f218a67b-55ef-4db3-bc55-e5a29fd47bb3

This happens because the file entries without `File` are being stripped away by this [function](https://github.com/arpitdalal/epic-stack/blob/232249c6478d590dc208b8bd681aab329a12b132/app/routes/users%2B/%24username_%2B/__note-editor.tsx#L55) and since the existing images are only being passed as `id` in a hidden field, they're being stripped away.
As a solution, I've updated that function not to strip away any images that have id, see [here](https://github.com/arpitdalal/epic-stack/blob/9f0452a133e500711cecbc33f4fb33e3eaebfa9f/app/routes/users%2B/%24username_%2B/__note-editor.tsx#L55) and update related code in `transform` and `prisma query`.
See this video after this fix

https://github.com/epicweb-dev/epic-stack/assets/61059807/a03728b2-5622-4b13-aec1-b0291047d296

Let me know if you need me to make any changes.

## Test Plan
I couldn't find any tests already written for this, so haven't updated anything.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/routes/users+/$username_+/__note-editor.tsx
+++ b/app/routes/users+/$username_+/__note-editor.tsx
@@ -52,10 +52,14 @@ const ImageFieldsetSchema = z.object({
 
 type ImageFieldset = z.infer<typeof ImageFieldsetSchema>
 
-function imageHasFile(
+function imageHasFileOrId(
 	image: ImageFieldset,
-): image is ImageFieldset & { file: NonNullable<ImageFieldset['file']> } {
-	return Boolean(image.file)
+): image is ImageFieldset &
+	(
+		| { file: NonNullable<ImageFieldset['file']> }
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thank you for this! I've adjusted the approach and I think it simplified things quite a bit. I appreciate you bringing this to my attention! I'm recording the video for EpicWeb.dev where we build this part of the app *right now* so it was very timely :)

PR #411, opened 2023-08-30
docs: Correct optional parameter syntax in caching code snippet

MoSattler:
Hi all, I've corrected a minor TypeScript syntax error in the caching.md document regarding the optional timings parameter.


kentcdodds (approved):
Thanks!

Issue #415, opened 2023-08-30
Cleanup after e2e tests properly


kentcdodds:
All the logic is there, but for some reason it's not working properly 🤷‍♂️

kentcdodds:
I recorded a screencast demonstrating the bug. Pretty sure it's a bug in playwright: https://www.loom.com/share/07253aa0e7114996ab7a2b226722ce41

Here's the AI summary:

> In this video, I discuss a problem we're facing with test cleanup in our codebase. I walk through the steps I took to identify the issue and explain the relevant code. Despite having proper cleanup mechanisms in place, we're experiencing failures when running multiple tests together. I demonstrate the problem and share my observations. By watching this video, you'll gain insight into the issue and hopefully help us find a solution.

kentcdodds:
This has now been fixed: https://github.com/epicweb-dev/epic-stack/commit/85465dbf83cfc70486043666c08270636b6dafb7

PR #420, opened 2023-08-31
add test cases for note images

arpitdalal:
Adding tests for note images.
The following tests are included:

1. Users can create note with an image
2. Users can create note with multiple images
3. Users can edit note image
4. Users can delete note image

## Checklist

- [x] Tests updated
- [ ] Docs updated

### DIFF
--- a/tests/e2e/note-images.test.ts
+++ b/tests/e2e/note-images.test.ts
@@ -1,0 +1,154 @@
+import fs from 'fs'
+import { faker } from '@faker-js/faker'
+import { expect, test } from '@playwright/test'
+import { type NoteImage, type Note } from '@prisma/client'
+import { prisma } from '#app/utils/db.server.ts'
+import { insertedUsers, loginPage } from '#tests/playwright-utils.ts'
+
+test('Users can create note with an image', async ({ page }) => {
+	const user = await loginPage({ page })
+	await page.goto(`/users/${user.username}/notes`)
+
+	const newNote = createNote()
+	const altText = 'cute koala'
+	await page.getByRole('link', { name: 'new note' }).click()
+
+	// fill in form and submit
+	await page.getByRole('textbox', { name: 'title' }).fill(newNote.title)
+	await page.getByRole('textbox', { name: 'content' }).fill(newNote.content)
+	await page.setInputFiles(
+		"input[name='images[0].file']",
+		'tests/fixtures/images/kody-notes/cute-koala.png',
+	)
+	await page.getByRole('textbox', { name: 'alt text' }).fill(altText)
+
+	await page.getByRole('button', { name: 'submit' }).click()
+	await expect(page).toHaveURL(new RegExp(`/users/${user.username}/notes/.*`))
+	await expect(page.getByRole('heading', { name: newNote.title })).toBeVisible()
+	await expect(page.getByAltText(altText)).toBeVisible()
+})
+
+test('Users can create note with multiple images', async ({ page }) => {
+	const user = await loginPage({ page })
+	await page.goto(`/users/${user.username}/notes`)
+
+	const newNote = createNote()
+	const altText1 = 'cute koala'
+	const altText2 = 'koala coder'
+	await page.getByRole('link', { name: 'new note' }).click()
+
+	// fill in form and submit
+	await page.getByRole('textbox', { name: 'title' }).fill(newNote.title)
+	await page.getByRole('textbox', { name: 'content' }).fill(newNote.content)
+	await page.setInputFiles(
+		"input[name='images[0].file']",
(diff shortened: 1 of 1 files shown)


tests/e2e/note-images.test.ts
```diff
@@ -0,0 +1,154 @@
...
+		},
+	} satisfies Omit<
+		Note,
+		'id' | 'createdAt' | 'updatedAt' | 'type' | 'ownerId'
+	> & {
+		images: { create: Pick<NoteImage, 'altText' | 'blob' | 'contentType'> }
+	}
+}
+
+test.afterEach(async () => {
```
kentcdodds, line 149: Is this necessary? We already handle this in the utils module I think. But it's not working so maybe this is necessary? #415 

I'd love to not require this

tests/e2e/note-images.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	const newNote = createNote()
+	const altText = 'cute koala'
+	await page.getByRole('link', { name: 'new note' }).click()
+
+	// fill in form and submit
+	await page.getByRole('textbox', { name: 'title' }).fill(newNote.title)
+	await page.getByRole('textbox', { name: 'content' }).fill(newNote.content)
+	await page.setInputFiles(
+		"input[name='images[0].file']",
```
kentcdodds, line 20: Is there another way to locate this input field? This feels like an implementation detail.

tests/e2e/note-images.test.ts
```diff
@@ -0,0 +1,154 @@
...
+		'tests/fixtures/images/kody-notes/cute-koala.png',
+	)
+	await page
+		.locator('[id="note-editor-images\\[0\\]\\.altText"]')
```
kentcdodds, line 48: Can we improve this locator?

kentcdodds (commented):
Thanks for working on this!

I would like to improve some of the locators and other things if possible.

tests/e2e/note-images.test.ts
arpitdalal: One solution is to use `getByLabel` but for multiple inputs, all of them have the same label `Image` which isn't ideal, so if the implementation is changed to have different labels for each image selection input then they can be selected using these unique labels.
Also, I think these multiple inputs should have different labels to make it easy for the screen readers.

Another solution is to not touch labels but have unique but predictable testids for each file input which can be selected using `getByTestId`.

Let me know how you want to approach this.

Screenshot 2023-09-01 at 10 31 36 AM

tests/e2e/note-images.test.ts
arpitdalal: `nth()` method can be used on a `getByLabel` method to select nth textarea, like so:
```js
await page
  .getByLabel(/Alt text/i)
  .nth(0) // selects 1st textarea with label "Alt text"
  .fill(altText1)
await page
  .getByLabel(/Alt text/i)
  .nth(1) // selects 2nd textarea with label "Alt text"
  .fill(altText2)
```
Other than this, we can change the label to be different for each textarea but not sure what it should be or if it will be helpful for other than testing 🤷‍♂️

tests/e2e/note-images.test.ts
arpitdalal: We can use the `nth` method on`getByLabel` just like I mentioned for textarea

tests/e2e/note-images.test.ts
arpitdalal: fixed in https://github.com/epicweb-dev/epic-stack/pull/420/commits/f13d773367c1f63ebc3eedc6808d0422277580a0

tests/e2e/note-images.test.ts
kentcdodds: Yes, I like that better.

arpitdalal: fixed in https://github.com/epicweb-dev/epic-stack/pull/420/commits/569c7a53fbc18b5fa8e15dfbb9d9ca89f8503d70

tests/e2e/note-images.test.ts
kentcdodds: Yeah, I'd like to completely avoid complicated locator strings. Could you do that here too?

arpitdalal: fixed here https://github.com/epicweb-dev/epic-stack/pull/420/commits/edae2f0f6efaa432b1d63963ebf3ddf502e8706c

kentcdodds:
Could you rebase your changes on `main`?

kentcdodds:
Thanks again! Could you revert changes to the package lock? I don't think that file should have changes

kentcdodds (approved):
Thanks!

PR #428, opened 2023-09-07
add tenant example

vinstah:
## Checklist

- [x] Docs updated

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -56,6 +56,9 @@ This page links to examples of how to implement some things with the Epic Stack.
 - [Epic Stack + Fathom Analytics](https://github.com/xstevenyung/epic-stack-with-fathom-analytics)
   by [@xstevenyung](https://github.com/xstevenyung): An example of the Epic Stack
   with Fanthom Analytics via CDN
+- [Epic Stack + Tenant Users](https://github.com/offseat/epic-stack-tenant)
+  by [@vinstah](https://github.com/vinstah): An example of the Epic Stack
+  with tenant users and members starter
   
 ## How to contribute
 


docs/examples.md
```diff
@@ -56,6 +56,9 @@ This page links to examples of how to implement some things with the Epic Stack.
 - [Epic Stack + Fathom Analytics](https://github.com/xstevenyung/epic-stack-with-fathom-analytics)
   by [@xstevenyung](https://github.com/xstevenyung): An example of the Epic Stack
   with Fanthom Analytics via CDN
+- [Epic Stack + Tenant Users](https://github.com/offseat/epic-stack-tenant)
```
kentcdodds, line 59: I think this may be private:

image


Could you make it public?

kentcdodds (commented):
This is great! Just one thing.

docs/examples.md
vinstah: doh, public now!

kentcdodds: Thanks!

Unfortunately, it's hard to tell what changes are required for this example. The instructions say you should make a commit as soon as you initialize the project and then after you make your changes so people can see the difference between a standard epic stack project and one with the changes you make.

Could you do that please?

vinstah: Hi, yeah I'll update that

vinstah: Hi Kent,

I've updated it now

kentcdodds: Thanks!

I'm concerned the implementation is overly complicated. What do you think about doing this instead:

1. create a tenant model with an id and name
2. give all models a `tenantId` and relate them to the tenant (one-to-many relationship)
3. adjust queries to scope them to a specific `tenantId` (may require updating the routes to include the `tenantId` in the params)

I think that would be simpler than having a `User`, `TenantUser`, `WorkspaceUser` and `Member` all of which kinda feel like the same abstract concept to me. What do you think?

vinstah: Yeah they do seem to be sort of the same at the moment, though I chose to add extraction so when I want to query a WorkspaceUser I can simply do `prisma.workspaceuser`, a TenantUser is usually the admin of the tenant and manages subscription/payment, Members are the Individuals that belong to the Tenant.

It's a complex system that does need thorough documentation

kentcdodds: Ok, sounds good. Thanks!

PR #431, opened 2023-09-08
Hide logs for images and Healthcheck

### LINKED ISSUE #368: Hide logs for images
They're not very useful and they're pretty noisy.

rperon:
If the response of the image or the healthcheck is ok, then we can skip them, otherwise a logs can be pretty useful.

I added the healthcheck path because they generate a lot of logs on deployment and only error healthcheck are interesting.

This should close #368 

## Test Plan

Disable network cache, and check the logs.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
*thank you*!!

I've been meaning to figure out how to do this.

PR #440, opened 2023-09-12
fix: some form id renamed to reflect the route it belongs to

natac13:
I noticed some form ids where all `signup-form`. I assumed this undesired and thought I would submit a small PR to help out.

Thanks for this great repo! 🙏 🚀

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Good call and nice catch 👍 Thanks!

PR #441, opened 2023-09-13
test: rework Playwright cleanup using fixtures

mxschmitt:
Hi Kent!

This is a PR regarding your recently filed issue in https://github.com/microsoft/playwright/issues/27058. We at Playwright recommend fixtures for the cleanup of resources, which make it super easy to cleanup things after the use. In this PR I moved some of your resources over to fixtures, so its a bit move obvious to you to understand what we described in the issue, let me know if you have any questions!

I saw your `UniqueEnforcer` instance, this might be problematic, since it does not sync across multiple Playwright workers. Also it might make sense to move `loginPage` over to a fixture, so it always has access to the `page` instance. Not sure about the design choices about the `User` type, I just created it for now, might be changed to type inference etc. later.

(Marked as a draft for now, since this is more a FYI / how we would do things PR)

### DIFF
--- a/tests/db-utils.ts
+++ b/tests/db-utils.ts
@@ -2,12 +2,16 @@ import fs from 'node:fs'
 import { faker } from '@faker-js/faker'
 import bcrypt from 'bcryptjs'
 import { UniqueEnforcer } from 'enforce-unique'
-import { getPasswordHash } from '#app/utils/auth.server.ts'
-import { prisma } from '#app/utils/db.server.ts'
+
+export type User = {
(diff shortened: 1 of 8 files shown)


kentcdodds:
Thanks for this @mxschmitt! This helped me try fixtures out again and I'm pretty happy with where I landed: https://github.com/epicweb-dev/epic-stack/commit/85465dbf83cfc70486043666c08270636b6dafb7

I really appreciate you taking the time to demonstrate this!

PR #442, opened 2023-09-13
Install packages with bun in github actions

dyeoman2:
(deploy.yml): Swap out npm with bun for faster package installation in the lint, typecheck, and vitest GitHub actions. Unable to swap out npm for bun in the playwright job due to a compatibility issue with better-sqlite3.

(vitest.config.ts): Replaced the defineConfig 'vite' import with the 'vitest/config' import because of a type issue that occurs when using bun with vite.

(auth..callback.test.ts, verify.tsx): Explicitly de-structured properties from generateTOTP because 'charSet' doesn't exists in the verification table but was appearing in '...config'.

## Test Plan

To ensure that the transition from NPM to Bun in the GitHub Actions workflow is successful, follow these steps:
1. Trigger a workflow run (either by pushing changes or creating a pull request to main/dev branches).
2. Monitor the GitHub Actions logs to verify that Bun is being installed and used without errors.
3. Check the individual jobs (lint, typecheck, vitest, playwright) in the workflow to ensure all steps are executed correctly.
4. Ensure that all the tests and checks pass without any errors.
5. After the tests, verify that any generated reports or logs (like the playwright report) are uploaded correctly.

## Checklist

- Bun is installed without any issues.
- Dependencies are downloaded using Bun.
- Scripts are executed with Bun without errors (like building icons, linting, type checking, etc.).
- Tests (vitest, playwright) are running and completing successfully.
- Playwright report is uploaded to the GitHub artifacts.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -27,8 +27,11 @@ jobs:
         with:
           node-version: 18
 
+      - name: 🍞 Install Bun
+        uses: oven-sh/setup-bun@v1
+
       - name: 📥 Download deps
-        uses: bahmutov/npm-install@v1
+        run: bun install
 
       - name: 🖼 Build icons
         run: npm run build:icons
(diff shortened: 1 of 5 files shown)


kentcdodds:
Thanks for this, but I don't think I feel comfortable testing differently than production.

PR #443, opened 2023-09-13
fixtures example

pavelfeldman:
This is just an FYI, I know you've already landed a fix, feel free to close it!


kentcdodds (commented):
Thank you for sharing this solution as well :)

PR #449, opened 2023-09-15
add i18n example

rperon:
i18n integration using i18next and remix-i18next. Language is detected by the browser, I added a dropdown menu to change the language and this value is saved inside a cookie. There is multiple way to save the user language, I started with cookie as it was easy but it can be change depending on feedback.
I was also wondering if I should create test in French, maybe the switch-language ?

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

Screenshot 2023-09-15 at 10 51 40


kentcdodds (approved):
This is great! Thank you!

PR #452, opened 2023-09-18
Removed --typescript flag, in remix v2 it is used as name for the directory

pr0gstar:
removed --typscriot flag because it is used as directory name with new remix version 2

## Checklist

- [x] Tests updated
- [x] Docs updated


kentcdodds (approved):
Thanks!

MichaelDeBoey (approved):

PR #453, opened 2023-09-18
fix typos

imcodingideas:
I fixed some typos as I was digging deep into this starter pack. There shouldn't be any regressions to consider.

## Test Plan

Should have no side effects, all tests should pass.

## Screenshots

n/a

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -40,7 +40,7 @@ import {
 	DropdownMenuTrigger,
 } from './components/ui/dropdown-menu.tsx'
 import { Icon, href as iconsHref } from './components/ui/icon.tsx'
-import fontStyleStyleSheetUrl from './styles/font.css'
+import fontStyleSheetUrl from './styles/font.css'
 import tailwindStyleSheetUrl from './styles/tailwind.css'
 import { authenticator, getUserId } from './utils/auth.server.ts'
 import { ClientHintCheck, getHints, useHints } from './utils/client-hints.tsx'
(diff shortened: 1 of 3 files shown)


app/root.tsx
```diff
@@ -40,8 +40,8 @@ import {
 	DropdownMenuTrigger,
 } from './components/ui/dropdown-menu.tsx'
 import { Icon, href as iconsHref } from './components/ui/icon.tsx'
-import fontStylestylesheetUrl from './styles/font.css'
-import tailwindStylesheetUrl from './styles/tailwind.css'
+import fontStyleStyleSheetUrl from './styles/font.css'
```
imcodingideas, line 43: I checked with MDN, and they refer to Stylesheet as StyleSheet. It felt wrong looking at it.

https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet

kentcdodds: Maybe it should be `fontStyleSheetUrl` 😅

imcodingideas: Yessir, I missed this with renaming haha.

app/routes/settings+/profile.tsx
```diff
@@ -1,12 +1,12 @@
 import { json, type DataFunctionArgs } from '@remix-run/node'
 import { Link, Outlet, useMatches } from '@remix-run/react'
+import { z } from 'zod'
```
imcodingideas, line 3: This was a eslint issue I saw in github

Screenshot 2023-09-18 at 9 07 36 AM

kentcdodds (approved):
Thanks!

PR #455, opened 2023-09-20
fix(entry.server): provide nonce to react scripts

nicholaschiang:
This patch provides the `nonce` option to the React `renderToPipeableStream` function so that React will add it to scripts dynamically injected (e.g. the `<Suspense>` inline scripts).

Ref: https://react.dev/reference/react-dom/server/renderToPipeableStream#parameters
Ref: https://github.com/facebook/react/issues/26026#issuecomment-1398640424
Ref: https://github.com/remix-run/remix/issues/5156#issuecomment-1397935995

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -72,6 +72,7 @@ export default async function handleRequest(...args: DocRequestArgs) {
 
 					console.error(error)
 				},
+				nonce,
 			},
 		)
 


kentcdodds:
Honestly not sure how I missed this 🙃

PR #456, opened 2023-09-20
Add robots.txt and sitemap.xml

lpsinger:
See https://github.com/remix-run/remix/discussions/2912#discussioncomment-7039395

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/routes/_seo+/sitemap[.]xml.ts
+++ b/app/routes/_seo+/sitemap[.]xml.ts
@@ -1,0 +1,12 @@
+import { generateSitemap } from '@nasa-gcn/remix-seo'
+import { routes } from '@remix-run/dev/server-build'
+import { type LoaderFunctionArgs } from '@remix-run/node'
+
+export function loader({ request }: LoaderFunctionArgs) {
+	return generateSitemap(request, routes, {
+		siteUrl: new URL(request.url).origin,
+		headers: {
+			'Cache-Control': `public, max-age=${60 * 5}`,
+		},
+	})
+}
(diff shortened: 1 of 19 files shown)


lpsinger:
I'm not sure what's going wrong when I run `npm run dev`:

```
$ npm run dev

> predev
> npm run build:icons --silent


> dev
> remix dev -c "node ./server/dev-server.js" --manual


 💿  remix dev

 info  building...
 info  built (8.4s)
6:30:10 PM [tsx] rerunning
Debugger listening on ws://127.0.0.1:9229/8911dd2a-b040-4bac-ac18-7207cb0b3a3c
For help, see: https://nodejs.org/en/docs/inspector
TypeError: Cannot read properties of undefined (reading 'startsWith')
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at file:///Users/lpsinger/src/epic-stack/tests/mocks/github.ts:1:2646
```

app/routes/_seo+/sitemap[.]xml.ts
```diff
@@ -0,0 +1,19 @@
+/*!
+ * Copyright © 2023 United States Government as represented by the
```
Comment (kentcdodds) [inline app/routes/_seo+/sitemap[.]xml.ts:2]: I understand why this has to be here, but I don't know how I feel about merging this with the copyright disclaimer.

Do I need to put together a CLA for the first time in my life? 😬

kentcdodds (commented):
Thanks for this! I can fix the issue you're having. But I'm concerned about something else.

Also, we probably should add some documentation.

app/routes/_seo+/sitemap[.]xml.ts
Comment (lpsinger) [reply app/routes/_seo+/sitemap[.]xml.ts:2]: Oops, I forgot to remove that.

kentcdodds (approved):
Thanks a bunch! This is very useful!

PR #459, opened 2023-09-22
docs: fix cli param

ananni13:
`--init` is not a valid option for `create-remix`.

Launching the script as it is now, the CLI will pick up `--init` as the folder name to use to create the project, and won't ask for one.

If the intention was to automatically launch the remix.init script, we should use `--init-script` instead.

## Checklist

~~- [ ] Tests updated~~
- [x] Docs updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 </div>
 
 ```sh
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thank you!

PR #472, opened 2023-10-02
move app name to ENV.APP_DISPLAY_NAME

sys13:
Moved app name (Epic Notes) to `ENV.APP_DISPLAY_NAME`. Haven't added docs yet but let me know if this is a good approach and I'll go do that.

(noticed ENV wasn't available in the tests but wasn't a big deal)

## Checklist

- [x] Tests updated
- [ ] Docs updated

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,7 @@ SESSION_SECRET="super-duper-s3cret"
 INTERNAL_COMMAND_TOKEN="some-made-up-token"
 RESEND_API_KEY="re_blAh_blaHBlaHblahBLAhBlAh"
 SENTRY_DSN="your-dsn"
+APP_DISPLAY_NAME="Epic Notes"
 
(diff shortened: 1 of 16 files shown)


kentcdodds:
Sorry for dragging my feet on this. The reason is I really prefer to avoid doing things like this generally. In the real apps I build, I never want something like this to be a config-based thing. Ultimately strings like this should go in localization strings, but in apps where I'm not localizing things I prefer to inline them rather than have environment variables.

I realize this is a bit of a pain when you're creating a new Epic Stack app to have to go through and update a bunch of hard-coded stuff, but I think it's better than leaving a bunch of environment variable references.

Sorry!

swalker326:
Could this be a use case for a code mod to update app name?

kentcdodds:
I thought about that, but it's really just in a bunch of strings and so a simpler find/replace thing would probably be better. I'm just concerned that it would be pretty brittle.

Issue #423, opened 2023-09-02
tests pass but throw errors after clean installation


rmobis:
I'm having a hard time getting the tests to run correctly after a clean install, getting some very cryptic error messages which seem related to mocking and/or Prisma (possibly two different issues?). These are the steps I've taken:

Run the setup script to initialize a new project and cd into it. When asked about setting up deployment, I skip it.

```bash
> npx create-remix@latest --typescript --install --template epicweb-dev/epic-stack epic-test
> cd epic-test
> npm run setup
> npm run test
```
This *kinda* works. All tests pass, but at the end instead of getting a summary I get this rust error:

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: failed to delete napi ref', query-engine\query-engine-node-api\src\engine.rs:138:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

<details>
<summary>Full Log</summary>
<pre><code> rmobi@DREAM-SHATTERER  C:      epic-test  npm run test

\> test
\> vitest

Environment variables loaded from .env
Prisma schema loaded from prisma\schema.prisma
Datasource "db": SQLite database "base.db" at "file:C:\projects\personal\epic-test\tests\prisma\base.db"

SQLite database base.db created at file:C:\projects\personal\epic-test\tests\prisma\base.db

Applying migration `20230811194017_init`

Database reset successful

The following migration(s) have been applied:

migrations/
  └─ 20230811194017_init/
    └─ migration.sql

Running seed command `tsx prisma/seed.ts` ...
🌱 Seeding...
🧹 Cleaned up the database...: 8.929ms
🔑 Created permissions...: 183.277ms
👑 Created roles...: 24.882ms
👍 Minimal seed complete
🌱 Database has been seeded: 218.381ms

The seed command has been executed.

 DEV  v0.34.2 C:/projects/personal/epic-test

 ❯ app/utils/misc.error-message.test.ts (3)
   ✓ Error object returns message
   ✓ String returns itself
   ✓ undefined falls back to Unknown
   \- [ afterAll ]
 ❯ app/routes/_auth+/auth.$provider.callback.test.ts (8) 397ms
   ✓ a new user goes to onboarding
   ✓ when auth fails, send the user to login with a toast
   ✓ when a user is logged in, it creates the connection
   ✓ when a user is logged in and has already connected, it doesn't do anything and just redirects the user back to the connections page
[...]

PR #479, opened 2023-10-08
Fix busy resource error in test's afterAll

embiem:
Fix the `Error: EBUSY: resource busy or locked, unlink 'hidden\path\tests\prisma\data.25504.db'` that occurs on Windows during the `afterAll` call in tests.

Related Issue: https://github.com/epicweb-dev/epic-stack/issues/423#issuecomment-1752018535

## Test Plan

- Ensure you're in a Windows environment
- Run `vitest --no-threads` ([ref issue](https://github.com/epicweb-dev/epic-stack/issues/423#issuecomment-1752017026) explaining why to use --no-threads)
- All tests pass & no `Error: EBUSY: resource busy or locked` for every `.db` created in `tests/prisma/`

### DIFF
--- a/tests/setup/db-setup.ts
+++ b/tests/setup/db-setup.ts
@@ -21,6 +21,6 @@ afterEach(async () => {
 
 afterAll(async () => {
 	const { prisma } = await import('#app/utils/db.server.ts')
-	prisma.$disconnect()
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #485, opened 2023-10-13
fix #484: export @epic-web/totp from server file to prevent Crypto polyfill

### LINKED ISSUE #484: Crypto from @epic-web/totp being included in the client build
I was trying to add pwa support to the epic-stack but got an error saying that `Crypto` is a node package and after some digging, I found out that node crypto from `@epic-web/totp` is being included in the client build as well.
To resolve this, I just exported everything from `@epic-web/totp` from a `.server.ts` file and imported everything from that `.server.ts` file instead of directly importing from `@epic-web/totp`

arpitdalal:
## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

PR #498, opened 2023-10-23
Bump tsx

xHomu:
Get rid of the --experimental-loader warning from node when using `tsx`:

image

 See https://github.com/esbuild-kit/tsx/issues/330

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -122,7 +122,7 @@
         "vitest": "^0.34.5"
       },
       "engines": {
-        "node": "^18 || ^20"
+        "node": "18"
       }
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Cool, thanks!

kentcdodds:
Looks like we need the lockfile to be updated.

xHomu:
Ah, good catch. 

Is there any reason why we're keeping node engine at 18?

xHomu:
I see, weird:

image

kentcdodds (approved):
Thanks!

PR #500, opened 2023-10-25
improve design progress bar

hakimLyon:
## Screenshots

https://github.com/epicweb-dev/epic-stack/assets/64654633/49ddf0d9-1a4f-4efb-9694-78d0f523af6f

### DIFF
--- a/app/components/progress-bar.tsx
+++ b/app/components/progress-bar.tsx
@@ -47,11 +47,11 @@ function EpicProgress() {
 				)}
 			/>
 			{delayedPending && (
-				<div className="absolute flex items-center justify-center">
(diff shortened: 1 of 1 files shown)


kentcdodds:
I'm not sure I agree this is better 🤔

swalker326:
I don't know if moving the spinner to the right is better. It's different and I think they both look fine. If a user wanted to move it they could, don't think it needs changed in the base project.

hakimLyon:
Got it. It's a matter of personal preference, and both options seem acceptable. Everyone has the choice to move it according to their preferences. I will close this PR.

PR #503, opened 2023-10-27
Added basic prometheus implementation 

swalker326:
Added base prometheus implementation to track metrics. 

## Test Plan

navigate to `localhost:9091/metrics` to see query events being tracks

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

Screenshot 2023-10-27 at 3 51 52 PM


server/prometheus.server.ts
```diff
@@ -0,0 +1,64 @@
...
+
+// Recreate the singleton function here, we don't want to import it from
+// the utils because this is used by server index.ts which can only import the JS
+// compiled directly from this file, and I don't want to move it to the server folder
+function singleton<Value>(name: string, value: () => Value): Value {
+    const g = global as any;
+    g.__singletons ??= {};
+    g.__singletons[name] ??= value();
+    return g.__singletons[name];
+}
```
kentcdodds, line 11: We can now use [`@epic-web/remember`](https://www.npmjs.com/package/@epic-web/remember) :)

server/index.ts
```diff
@@ -92,10 +98,20 @@ app.use(
 // more aggressive with this caching.
 app.use(express.static('public', { maxAge: '1h' }))
 
-app.get(['/build/*', '/img/*', '/fonts/*', '/favicons/*'], (req, res) => {
-	// if we made it past the express.static for these, then we're missing something.
-	// So we'll just send a 404 and won't bother calling other middleware.
-	return res.status(404).send('Not found')
```
kentcdodds, line 98: Could you bring this back? It's a recent addition I'm guessing you didn't have when you forked 😅

server/index.ts
```diff
@@ -20,6 +20,12 @@ import rateLimit from 'express-rate-limit'
 import getPort, { portNumbers } from 'get-port'
 import helmet from 'helmet'
 import morgan from 'morgan'
+import { type prometheus as prometheusType } from '#server/prometheus.server.ts'
+// @ts-ignore - this file may not exist if you haven't built yet, but it will
+// definitely exist by the time the dev or prod server actually runs.
+import { prometheus as _prometheus } from '#server-build/prometheus.server.js'
```
kentcdodds, line 26: I'm pretty sure you should be able to do this:

```
import { prometheus } from './prometheus.server.ts'
```

And that should get compiled properly. If not, I'm pretty sure there's something we can change in our build to make it work.

app/utils/db.server.ts
```diff
@@ -34,3 +36,9 @@ export const prisma = remember('prisma', () => {
 	client.$connect()
 	return client
 })
+
+const recordMetrics = (e: Prisma.QueryEvent) => {
+	prometheus.sqlQueryCounter.inc({ query: e.query })
+	prometheus.sqlQueryDuration.observe({ query: e.query }, e.duration)
+	prometheus.sqlQueryDurationSummary.observe({ query: e.query }, e.duration)
+}
```
kentcdodds, line 44: Since this is only called in one place I think I'd rather inline it above instead of have this function.

kentcdodds (changes requested):
Thanks! This is a great start! I have a few comments. Also, would you mind adding a new docs markdown file for this?

server/prometheus.server.ts
swalker326: replaced with remember.

swalker326:
> Thanks! This is a great start! I have a few comments. Also, would you mind adding a new docs markdown file for this?

Awesome, I resolved all the comments but I'm having an issue with the playwright run. I'll have more time this week to look at it, but if anyone wanted to take a look in the mean time. I'm fairly certain the issue revolves around  how the server-build is importing the TS version of the prometheus.server.ts, in the JS file built file its trying to import `prometheus.server.ts`. 

I will for sure work on getting some kind of docs updated too.

kentcdodds:
Sorry about that! I think this should work :)

swalker326 (commented):
I think I addressed all changes

server/index.ts
```diff
@@ -216,6 +233,31 @@ app.all(
...
+const metricsServer = metricsApp.listen(metricsPort, () => {
+	if (metricsPort !== desiredMetricsPort) {
+		console.warn(
+			chalk.yellow(
+				`⚠️  Metrics port ${desiredMetricsPort} is not available, using ${metricsPort} instead.`,
+			),
+		)
+	}
+	console.log(`📈 Metrics server listening on port ${metricsPort}`)
+})
```
kentcdodds, line 259: Should we only enable this in production somehow? I'm thinking an environment variable `ENABLE_METRICS` or something. Most of the time local development really doesn't need the metrics.

swalker326: Added a environment variable, and checked it and production before launching the metrics server.

server/index.ts
```diff
@@ -98,6 +100,22 @@ app.get(['/build/*', '/img/*', '/fonts/*', '/favicons/*'], (req, res) => {
...
+			method: req.method,
+			path: req.path,
+			status: res.statusCode.toString(),
+		})
+		prometheus.requestDuration.observe({ path: req.path }, duration)
+		prometheus.requestDurationSummary.observe({ path: req.path }, duration)
+	})
+
+	next()
+})
```
kentcdodds, line 117: Could probably put this in an if block as well

docs/metrics.md
```diff
@@ -0,0 +1,18 @@
...
+
+Prometheus is an open-source systems monitoring and alerting toolkit. It's known for its reliability and scalability, making it a top choice for monitoring in modern cloud-native environments.
+
+### SQL Query Metrics (built in)
+
+Epic Stack apps provide:
+- **Count**: Number of SQL queries executed.
+- **Duration**: Time taken for each SQL query.
+
+Access the Prometheus dashboard for real-time metrics on port 9091
```
kentcdodds, line 15: Let's also mention the environment variable for enabling metrics.

kentcdodds (changes requested):
This is super! Just a couple last things to tidy up and we're good to go. Thank you!

server/index.ts
```diff
@@ -265,6 +312,9 @@ closeWithGrace(async () => {
 	await new Promise((resolve, reject) => {
 		server.close(e => (e ? reject(e) : resolve('ok')))
 	})
+	await new Promise((resolve, reject) => {
+		metricsServer.close(e => (e ? reject(e) : resolve('ok')))
```
kentcdodds, line 316: Does this cause a problem if we never call listen on this server? Maybe this should be in the if as well?

kentcdodds (commented):
Sorry, just realized one more thing

server/index.ts
swalker326: It doesn't case a problem, but I can wrap it still.

swalker326: Actually I didn't like it so I took a bit different approach. I know we were on the goal line, but this feels cleaner than if's all over checking the .env

kentcdodds:
Hey @swalker326, I just had a [conversation with Milin](https://twitter.com/kentcdodds/status/1719828873117151439) (the CEO of Sentry) on something unrelated to this. But then this effort came up and Milin told me that Sentry has a lot of really awesome features that we can use to accomplish the same thing but better. I hope you don't mind if I pump the breaks on this until I can meet up with someone else from Sentry to evaluate what their offering is.

I always hate doing this kind of thing because I know you and @matt-winfield put a lot of effort into contributing these improvements to the stack. I hope you understand as we work hard to come up with the very best solution for everyone.

matt-winfield:
Excited to hear about what Sentry can bring to this area! More options are always a good thing.

And I just like to share the things I was going to be investigating anyway, hearing about cool alternatives is one of the main benefits - doesn't really matter to me if my original idea ends up being included 🙂

swalker326:
I share the same sentiment as @matt-winfield. No worries here at all, was fun to dig into it.

jjaybrown:
Would @matt-winfield @swalker326 mind sharing the grafana dashboard which uses these metrics please?

kentcdodds:
Now that sentry is well supported,I think we can choose this

kentcdodds:
Thanks so much for the work on it though!

PR #505, opened 2023-10-29
Increase prettier printWidth to 100 (except for md files)

kasperpeulen:
I increased the printWidth to 100, except for md files.

I kept it to 80 for md files as a line length of 80 or even shorter is ideal for reading sentences:
https://baymard.com/blog/line-length-readability

However, in my opinion this is not the case with reading code, and it is more important, that doing "1 thing" can fit in "1 line", so that the length of the code gets a better indication of its actual complexity:

image

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -4,11 +4,7 @@ const appFiles = ['app/**']
 
 /** @type {import('@types/eslint').Linter.BaseConfig} */
 module.exports = {
-	extends: [
-		'@remix-run/eslint-config',
-		'@remix-run/eslint-config/node',
-		'prettier',
-	],
+	extends: ['@remix-run/eslint-config', '@remix-run/eslint-config/node', 'prettier'],
 	rules: {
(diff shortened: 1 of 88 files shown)


swalker326:
I would argue this makes reading code more difficult. I read a lot of git diffs and scrolling horizontally can be very frustrating imo. I would rather keep the template at 80, which is what prettier defaults to and add clear directions on how to change it in the docs. Just my thoughts, I'm not strongly opinionated about it either way.

MichaelDeBoey:
I would say we should stick to Prettier's default
I see them as having the most expertise on this topic and my opinion is they chose it with a reason (and have a good explanation for it in their docs)

This is imo also why we choose an opinionated code formatter: to not have to bother with personal preferences

kentcdodds:
Personally I like printWidth of 80. I like that it encourages me to reduce the amount of indentation I'm doing on my code.

We're going to stick with the default here. Thanks though!

PR #506, opened 2023-10-30
Add 404 assertion to e2e error-boundary

SanderKoopmans:
I noticed a `// todo: figure out` which seemed a nice little challenge. Add a listener to verify the 404

## Test Plan

- Checkout branch
- Run e2e tests

## Checklist

- [ x ] Tests updated
- [ n/a ] Docs updated

### DIFF
--- a/tests/e2e/error-boundary.test.ts
+++ b/tests/e2e/error-boundary.test.ts
@@ -1,8 +1,17 @@
+import { type Request } from '@playwright/test'
 import { expect, test } from '#tests/playwright-utils.ts'
 
 test('Test root error boundary caught', async ({ page }) => {
-	await page.goto('/does-not-exist')
+	const pageUrl = '/does-not-exist'
+	await page.goto(pageUrl)
 
 	await expect(page.getByText(/We can't find this page/i)).toBeVisible()
-	// TODO: figure out how to assert the 404 status code
+
+	const listener = (request: Request) => {
+		if (request.url().includes(pageUrl)) {
+			request.response().then(response => expect(response?.status()).toBe(404))
+		}
+	}
+
+	page.on('request', listener)
 })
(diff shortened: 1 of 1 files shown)


swalker326:
Thats clever, well done.

tests/e2e/error-boundary.test.ts
```diff
@@ -1,8 +1,17 @@
...
-	await page.goto('/does-not-exist')
+	const pageUrl = '/does-not-exist'
+	await page.goto(pageUrl)
 
 	await expect(page.getByText(/We can't find this page/i)).toBeVisible()
-	// TODO: figure out how to assert the 404 status code
+
+	const listener = (request: Request) => {
+		if (request.url().includes(pageUrl)) {
+			request.response().then(response => expect(response?.status()).toBe(404))
```
kentcdodds, line 12: With assertions that happen asynchronously like this I always like to make certain that they can fail. I'm concerned that as it's currently written, playwright will think the test is over before this assertion runs. Can you verify that the test can fail? https://kentcdodds.com/blog/make-your-test-fail

kentcdodds (commented):
Thanks! Just one thing.

SanderKoopmans:
Ha excellent, I'll see if I can break the test tomorrow.

caprolactam:
I write like this in my code. Please share what is the problem here.
```ts
test('Test root error boundary caught', async ({ page }) => {
  const res = await page.goto('/does-not-exist')

  expect(res?.status()).toBe(404)
  await expect(page.getByText(/We can't find this page/i)).toBeVisible()
})
```
[ref](https://playwright.dev/docs/api/class-page#page-goto-return)

kentcdodds:
Ah, yes I definitely prefer that!

SanderKoopmans:
Not sure if/why I didn't try that, suggestion applied. Another todo done ✅

kentcdodds:
Super! And have you verified that it can fail?

SanderKoopmans:
It fails as soon as someone changes the return status in the loader from `routes/$.tsx`. Second part fails as soon as someone changes the statusHandlers from the error boundary.

Is there anything else I should verify? ☺️

kentcdodds:
Great! That's enough. Thank you!

PR #507, opened 2023-10-31
fix: handling when session exists but user doesn't exist in db

caprolactam:
change handling from `authenticator.logout()` to `logout()`.

**Why**
 This process is intended to remove invalid session about userId on both cookie and db sides, when userId which get from session doesn't exists in db.
 We manage userId in `authSessionStorage` now. On the other hand, authenticator handle connection, which is stored in `connectionSessionStorage`. Therefore, even after `authenticator.logout()`, the information about the userId remains. This can cause infinite redirect. I think this is left over from the process when remix-auth and remix-auth-form were used for authentication.
 Replace `authenticator.logout()` with `logout()`, but if necessary, I will add the process to destory connectionSession manually.
For example:
```tsx
import { connectionSessionStorage } from '#app/utils/connections.server.ts'
// in action code.
const connectioSession = await connectionSessionStorage.getSession(
  request.headers.get('cookie'),
)
const responseInit = {
  headers: { 'set-cookie': connectionSessionStorage.destroySession(connectioSession) }
}
await logout({ request }, responseInit)
```

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -45,7 +45,7 @@ import {
 import { Icon, href as iconsHref } from './components/ui/icon.tsx'
 import fontStyleSheetUrl from './styles/font.css'
 import tailwindStyleSheetUrl from './styles/tailwind.css'
-import { authenticator, getUserId } from './utils/auth.server.ts'
+import { getUserId, logout } from './utils/auth.server.ts'
 import { ClientHintCheck, getHints, useHints } from './utils/client-hints.tsx'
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Wow! Thank you for finding this and fixing it!

I think your analysis is correct. This is left over code. I'm glad you've fixed it.

PR #509, opened 2023-10-31
Use svg over png for logos

kasperpeulen:
Use svg over pngs for the logos to save some bytes that need to be downloaded.


kentcdodds (approved):
Cool. Thanks!

PR #514, opened 2023-11-06
fix: `json` import

samialdury:
This PR fixes the import of `json` helper function from Remix, which should come from `@remix-run/node` to maintain consistency with all the other files.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thank you!

PR #515, opened 2023-11-07
feat(vscode): prompt to use TS from workspace

samialdury:
This PR adds two new settings to the VS Code workspace settings.

**Without these settings, VS Code uses its internal TS version, which can differ from the workspace. This could result in different errors/warnings shown in the editor and `npm run typecheck`.**

1. `typescript.tsdk`
  - VS Code setting description: Specifies the folder path to the tsserver and lib*.d.ts files under a TypeScript install to use for IntelliSense, for example: ./node_modules/typescript/lib. **When specified as a workspace setting, typescript.tsdk allows you to switch to use that workspace version of TypeScript for IntelliSense with the TypeScript: Select TypeScript version command.**
2. `typescript.enablePromptUseWorkspaceTsdk`
  - VS Code setting description: **Enables prompting of users to use the TypeScript version configured in the workspace for Intellisense.**
  - Related to the `typescript.tsdk` option, this results in this popup when opening the workspace: 
Screenshot 2023-11-07 at 12 34 31

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,5 +15,7 @@
   ],
   "workbench.editorAssociations": {
     "*.db": "sqlite-viewer.view"
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true
 }


.vscode/settings.json
```diff
@@ -15,5 +15,7 @@
   ],
   "workbench.editorAssociations": {
     "*.db": "sqlite-viewer.view"
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true
```
kentcdodds, line 20: I'm not sure I want to enable this. I'm pretty sure we always want people to use the built-in tsdk if it exists and a prompt would be annoying I think.

kentcdodds (changes requested):
Thanks! Just one thought.

.vscode/settings.json
samialdury: @kentcdodds This is the only way to enable it due to security, as per [this answer on StackOverflow](https://stackoverflow.com/a/75357623):
Screenshot 2023-12-10 at 12 44 19

**If you don't like this, I understand and feel free to close this PR!**

kentcdodds:
I'll need to try it out myself first. I'll get back to you.

.vscode/settings.json
kentcdodds: I just realized I already have the `"typescript.tsdk": "node_modules/typescript/lib",` setting enabled in my config, but I don't have `"typescript.enablePromptUseWorkspaceTsdk": true` enabled. So is this saying that my `typescript.tsdk` setting is just not doing anything? 🤔

PR #519, opened 2023-11-09
use ubuntu-22.04 for CI

fdietze:
Using `ubuntu-latest` is a bit of a gamble, as it can change over time. This commit switches it to `ubuntu-22.04` to make ci more reproducible.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Good call! I'm also considering using the `container.image` config so we use the same image as our production environment. I'll go ahead and merge this right now though. Thanks!

PR #523, opened 2023-11-11
Rese getPasswordHash function in resetUserPassword

JulesBlm:
Reuse the `getPasswordHash` function in `resetUserPassword` instead of using `bcrypt.hash(password, 10)` directly 

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks! That is definitely an oversight.

PR #526, opened 2023-11-16
feat(config): use SITE_NAME constant

shynonagons:
`Epic Notes` site name is hardcoded in a bunch of places, updated to a constant for better initial configurability. 

## Checklist

- [x] Tests updated
- [x] Docs updated

### DIFF
--- a/app/config/index.ts
+++ b/app/config/index.ts
@@ -1,0 +1,1 @@
+export const SITE_NAME='Epic Notes'
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -43,6 +43,7 @@ import {
 	DropdownMenuTrigger,
 } from './components/ui/dropdown-menu.tsx'
 import { Icon, href as iconsHref } from './components/ui/icon.tsx'
+import { SITE_NAME } from './config/index.ts'
 import fontStyleSheetUrl from './styles/font.css'
 import tailwindStyleSheetUrl from './styles/tailwind.css'
 import { getUserId, logout } from './utils/auth.server.ts'
@@ -90,7 +91,7 @@ export const links: LinksFunction = () => {
 
 export const meta: MetaFunction<typeof loader> = ({ data }) => {
 	return [
-		{ title: data ? 'Epic Notes' : 'Error | Epic Notes' },
+		{ title: data ? SITE_NAME : `Error | ${SITE_NAME}` },
 		{ name: 'description', content: `Your own captain's log` },
 	]
 }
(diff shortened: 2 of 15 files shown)


kentcdodds:
Thanks. This has been brought up before and I don't want to do anything with this until we've determined the mechanism for localizing strings. I understand it's a pain to find/replace all the places the title of the app appears when you start a new app, but I don't know of a single real-world app that uses an environment variable or a config like this for the title of the app. So I'd rather have the find/replace approach in place until we get around to localization. Thanks

dalelotts:
I can sort of see your reasoning, but this is a huge pain, especially when the dev is not familiar with the code base. 

Why not, at least for now, have vite statically replace the values at build time? https://vitejs.dev/guide/env-and-mode.html

kentcdodds:
For the same reason. 

The real solution is implementing internationalization. Until then it's find/replace. I'm open to docs on what to replace and even making it a part of the setup script is anyone's interested in doing that.

dalelotts:
I disagree with this decision but its your project - progress over perfection - it 15 minutes to put the variables in place and update GH secrets.

If others come across this, it's a SUPER simple one-time change:

Create ./types/vite-env.d.ts with the following content:

```ts
/// <reference types="vite/client" />

interface ImportMetaEnv {
	readonly VITE_APP_NAME: string
	readonly VITE_APP_DESCRIPTION: string
}

interface ImportMeta {
	readonly env: ImportMetaEnv
}
```

The you can update meta functions to use the variable. e.g. This is from root.tsx

```tsx
export const meta: MetaFunction<typeof loader> = ({ data }) => {
	return [
		{
			title: data
				? import.meta.env.VITE_APP_TITLE
				: `Error | ${import.meta.env.VITE_APP_TITLE}`,
		},
		{ name: 'description', content: import.meta.env.VITE_APP_DESCRIPTION },
	]
}
```

Or signup.tsx

```tsx
	const response = await sendEmail({
		to: email,
		subject: `Welcome to ${import.meta.env.VITE_APP_TITLE}!`,
		react: <SignupEmail onboardingUrl={verifyUrl.toString()} otp={otp} />,
	})
```


Or forgot-password.tsx

```tsx
	const response = await sendEmail({
		to: user.email,
		subject: `${import.meta.env.VITE_APP_TITLE} Password Reset`,
		react: (
			<ForgotPasswordEmail onboardingUrl={verifyUrl.toString()} otp={otp} />
		),
	})
```

PR #527, opened 2023-11-17
Upgrade pkgs in package.json to latest

rajeshg:
Update all packages in package.json to latest. Is there anything that we can do to make better use of the vite features in epic-stack?

## Test Plan

`npm run test` works as expected.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
This is awesome. Thanks!

PR #529, opened 2023-11-18
Fixed min-width on forms for mobile

tanmay-pathak:
## Summary

Updated the current `min-w-[368px]` to now apply on small screens and above. This was causing the form fields to overflow on mobile.

## Test Plan
https://epicstack.dev/signup can be visited on mobile to view the bug and then the fix can be viewed to compare if the issue has been resolved.

## Checklist

- ❌ Tests updated
- ❌ Docs updated

## Screenshots

|Bug|Fix|
|---|---|
|image|image|


kentcdodds (approved):
Thanks!

PR #533, opened 2023-11-21
Use "useActionData" in note editor

itsmegood:
Using noteFetcher.Form instead of just Form actually get the data back from the action.

P.s. I am not sure if this is done intentionally, if it is I would love to know the reason.

### DIFF
--- a/app/routes/users+/$username_+/__note-editor.tsx
+++ b/app/routes/users+/$username_+/__note-editor.tsx
@@ -17,7 +17,7 @@ import {
 	type DataFunctionArgs,
 	type SerializeFrom,
 } from '@remix-run/node'
-import { Form, useActionData } from '@remix-run/react'
+import { Form, useActionData, useNavigation } from '@remix-run/react'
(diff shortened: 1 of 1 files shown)


kentcdodds (commented):
Thanks! There is a mistake here. However this form is a navigation (we go somewhere else after successful submission and there can only be one at a time) so we really should switch this to `useActionData` and `useNavigation` for determining the pending state. Would you be interested in making that update?

itsmegood:
Hi! Kent hope you are doing great! I removed `useFetcher` with `useActionData` also I am confused on the usage of `useNavigation` because we are return status from the action, do we remove that code.

Sorry if I am being unhelpful here

itsmegood:
I submitted both of the branches have a take a look the first one is just using `useActionData` and a hacky way to get the loading or isPending state the second one is using `useNavigation` I will leave the choice up to you.

kentcdodds (approved):
Thank you!

PR #541, opened 2023-11-30
properly ignore healthcheck related transactions to sentry

lamson:
When reading 'x-healthcheck' from Sentry event request headers, we can't use `headers.get()` which is case-insensitive. Hence we'd need to use lowercase `x-healthcheck` key to properly check the request header value.

Also when the app is deployed to fly.io, there are other healthcheck events (from Consul) which we need to stop sampling.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

Before
Screenshot 2023-11-29 at 4 54 51 PM

After
Screenshot 2023-11-29 at 4 55 49 PM


kentcdodds (approved):
Thank you!

PR #542, opened 2023-11-30
fix (server): Fix source map retrieval for updated builds

kiliman:
This PR adds the fix submitted in Remix PR #8174. The issue is caused when the build server imports an updated server build with a ?t=timestamp suffix. This causes the source map support library to fail to load the source map.

https://github.com/remix-run/remix/pull/8174

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 import 'dotenv/config'
-import 'source-map-support'
+import ssourceMapSupport from 'source-map-support'
 import { installGlobals } from '@remix-run/node'
 import chalk from 'chalk'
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,9 +1,23 @@
 import 'dotenv/config'
-import 'source-map-support/register.js'
+import 'source-map-support'
```
kentcdodds, line 2:
```suggestion
import ssourceMapSupport from 'source-map-support'
```

kiliman:
Yeah, sorry about that. I just did a copy+paste from the Express template and assumed yours was the same. I should have run the lint/test scripts.

This should pass now.

PR #548, opened 2023-12-06
bugfix: set theme cookie maxAge so it's not a session cookie

xHomu:
Without max age, the cookie would disappear after the current session.


kentcdodds (approved):
Thanks!

PR #550, opened 2023-12-07
Add error boundary for edit note route

rajeshg:
1. Added an error boundary so that '/users/kody/notes/<UNKNOWN ID>/edit' results in the correct response. Without this user is shown the following error message - 'No user with the username "kody" exists'
**Before:**
image

**After:**
image

2. Updated packages to the latest (vite, prisma etc)

## Checklist

- [x] Tests updated
- [ ] Docs updated


kentcdodds (changes requested):
Thanks for this! Your PR updates all the dependencies and I prefer to do that in a separate PR or by my self. Could you reduce the things this PR does to one thing?

rajeshg:
Reverted changes to package.json

kentcdodds (approved):
Thank you!

PR #553, opened 2023-12-08
Fix Tiny Spelling Error

jjhiggz:
Fixed spelling error, it said "suing" which is a valid word, but should be "using"

## Checklist

- [ ] Tests updated
- [X] Docs updated

## Screenshot

See Code changed (It's one word)


kentcdodds (approved):
Thanks!

PR #558, opened 2023-12-12
feat(.vscode): add remix.code-snippets

hakimLyon:
## Screenshots

https://github.com/epicweb-dev/epic-stack/assets/64654633/6543eb0b-6505-4bb5-9704-f2931b33b43d

### DIFF
--- a/.vscode/remix.code-snippets
+++ b/.vscode/remix.code-snippets
@@ -1,0 +1,76 @@
+{
+	"loader": {
+		"prefix": "/ loader",
+		"body": [
+			"import { type DataFunctionArgs, json } from \"@remix-run/node\"",
+			"",
+			"export async function loader({ request }: DataFunctionArgs) {",
+			"  return json({})",
+			"}"
+		]
+	},
+	"action": {
+		"prefix": "/action",
+		"body": [
+			"import { type DataFunctionArgs, json } from \"@remix-run/node\"",
+			"",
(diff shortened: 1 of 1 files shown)


.vscode/remix.code-snippets
```diff
@@ -0,0 +1,76 @@
+{
+	"loader": {
+		"prefix": "/ loader",
```
kentcdodds, line 3:
```suggestion
		"prefix": "/loader",
```

kentcdodds (commented):
Thanks! Is there a way to limit these snippets to only apply in ts/js/tsx/jsx files?

.vscode/remix.code-snippets
hakimLyon: Yes. In Visual Studio Code, for example, you can do this by adding a "scope" field with the desired language identifiers. 
```json
		"scope": "typescriptreact,javascriptreact,typescript,javascript",
```

hakimLyon:
> Thanks! Is there a way to limit these snippets to only apply in ts/js/tsx/jsx files?

Yes. In Visual Studio Code, for example, you can do this by adding a "scope" field with the desired language identifiers. 
```json
		"scope": "typescriptreact,javascriptreact,typescript,javascript",
```

kentcdodds (approved):
Awesome! Thanks 🙏

PR #563, opened 2023-12-15
feat: add `autocomplete` prop to inputs

samialdury:
Added [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) prop to relevant inputs.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Love it. Thanks!

PR #566, opened 2023-12-16
Remove redundant and non-existant `border-border`

fdietze:
`border-border` doesn't exist, this was probably meant to be `border-box` (https://tailwindcss.com/docs/box-sizing#including-borders-and-padding).

But even that would be redundant, since it is already the default in the preflight syles.

### DIFF
--- a/app/styles/tailwind.css
+++ b/app/styles/tailwind.css
@@ -80,9 +80,6 @@
 }
 
 @layer base {
-	* {
-		@apply border-border;
-	}
 	body {
 		@apply bg-background text-foreground;
 	}


kentcdodds:
Thanks. But it actually does exist in the tailwind config: https://github.com/epicweb-dev/epic-stack/blob/b641031a5165e4323d67abb1290df89e98312741/app/utils/extended-theme.ts#L5

fdietze:
Oh, sorry, I see. Why is it useful to have a border color on every single html element?

kentcdodds:
It's useful to have the border color set so when you add a border you don't need to worry about setting the color as well

PR #574, opened 2023-12-26
fix: stop removing .git on init

jacobparis:
Discussed on Discord https://discord.com/channels/715220730605731931/1105271172079300630/1172436921427955743

My usual workflow for creating new projects is to spin up a new repo on Github, clone it to my machine, and then initialize whatever project I have inside it 

The epic stack currently deletes the .git directory in this case 

Another risk factor is if a dev decides to create a new branch of an existing project and runs the setup command, it will delete their .git and all their other branches

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

PR #575, opened 2023-12-27
fix: move border color to tailwind config

jacobparis:
The tailwind config is a better place to put this global color setting for the borders, especially since `.border-border` isn't a very descriptive class

I would also be in favour of removing `border` from the custom colors entirely as I think this is the only place that's used, but that might be a breaking change in case other folks have been using it 

```diff
	colors: {
- 		border: 'hsl(var(--border))',
		input: {
			DEFAULT: 'hsl(var(--input))',
			invalid: 'hsl(var(--input-invalid))',
```

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots
Before this change
image

After this change

image

### DIFF
--- a/app/styles/tailwind.css
+++ b/app/styles/tailwind.css
@@ -78,12 +78,3 @@
 		--ring: 217.2 32.6% 60%;
 	}
 }
-
-@layer base {
-	* {
-		@apply border-border;
-	}
-	body {
-		@apply bg-background text-foreground;
-	}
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thanks!

There are no breaking changes in the Epic Stack. Just thoughtful improvements to the starter with announcements for the big stuff.

PR #582, opened 2024-01-06
Marketing homepage redesign

simonswiss:
Redesign of the Epic Stack Marketing homepage, which corresponds to the page we'll build in one of the segments of the "Unleash The Designer In You" Epic Web workshop.

https://github.com/epicweb-dev/epic-stack/assets/485747/2e4ca359-8ec3-45be-ae62-e5e9df401289

Animations have been revisited and implemented with CSS only — no more GSAP or Framer Motion 👍

The design supports dark mode.


app/utils/extended-theme.ts
```diff
@@ -99,9 +99,24 @@ export const extendedTheme = {
 			from: { height: 'var(--radix-accordion-content-height)' },
 			to: { height: '0' },
 		},
+		'roll-reveal': {
```
kentcdodds, line 102: We want to make it easy for people to remove unused stuff if they want to. Having this code isn't really all that harmful, but I think it would be best if we add a comment here about what they can delete or even better if we can put these animation keyframes in the marketing index page and import them here. That way when people delete that page, TypeScript can tell them to delete the reference here. Perhaps we only need to have the marketing index page export the keyframes names and this imports those names for use here. That way there's a strong relationship between those keyframes and where they're used. We just want to make sure it's easy for people to know what to remove when they bootstrap a new Epic app.

To be clear, you don't have to do the same in the workshop. It's fine to keep it the way you have it for the workshop.

kentcdodds (commented):
This is terrific! I just have one thought

simonswiss:
See how you feel about this [new commit](https://github.com/epicweb-dev/epic-stack/pull/582/commits/5dfff0b4b389379d0265b55b78aba534cf258993).

The keyframes are defined as a Tailwind `preset` inside the `_marketing` route folder (I was not able to export it directly from the `_marketing+/index.tsx` route).

Since there is already a logos folder with loads of SVG in there, I assumed that was a safe place to put something that might be deleted by the user.

If deleted, the Tailwind config (which imports that preset) will flag that the `preset` doesn't exist and should be removed.

kentcdodds (approved):
I'm even happier with this. Thanks!

kentcdodds:
I'm not sure about the error you had, but I think you're someone was good 👍

kentcdodds:
Just occurred to me to look at the dark mode version and I think we could improve the logos a fair bit for that. Maybe just invert?

Screenshot_20240106_225629_Brave

simonswiss:
Whoops yes! I did something wrong — this is how it was supposed to look. I put the `grayscale` on the parent and it grayscaled the tiles as well:

https://github.com/epicweb-dev/epic-stack/assets/485747/63264cdb-426a-4c96-8958-a255d4693823

The idea of grayscale is the page is slightly less colorful, but the logos still show their colors on hover 👍

But I am not convinced it works without changing the logo colors — so maybe I'll just go with lighter tiles on dark.

Will open a new PR 👍

Issue #583, opened 2024-01-06
Admin protected routes are not showing relevant error message.


aerojeyenth:
**Steps to reproduce:**

1. Login to epic stack with a non-admin user 
2. Navigate to ./admin/cache route
3. Observe the error message 

**Actual behavior:** 

Error message is undefined
`Cannot read properties of undefined (reading 'message')`

**Expected behavior:**

Relevant error message:

`Unauthorized: required role: admin`

Please find the screenshot attached. 

![image](https://github.com/epicweb-dev/epic-stack/assets/6432625/72712eec-d307-4af9-b393-3c4b473e86a0)

kentcdodds:
Good catch. Would you be interested in submitting a PR to fix this?

aerojeyenth:
> Good catch. Would you be interested in submitting a PR to fix this?

Sure. ! Do you want to give me any pointers ? My initial thought is fix is needed in the <GeneralErrorBoundary />

kentcdodds:
I'd have to dig in a little bit. You should probably start in that route and trace the code.

paolapog:
Hello @kentcdodds and @aerojeyenth,  I've found a way to solve this issue, I don't know if the most correct one but let me know, at least we have some clue :) 
cheers

kentcdodds:
Thanks!

PR #596, opened 2024-01-12
fix: add general error boundary on cache routes

paolapog:
I've found a way to solve #583 
Basically was missing a `<GeneralErrorBoundary> `component with its error handling component because if you were taking the error from the cache routes was resulting in an undefined result.
The error from general error boundary is an object with inside data that contains the error.
## Test Plan
1. Login to epic stack with a non-admin user
2. Navigate to ./admin/cache route
3. See the correct error being showed
Screenshot 2024-01-12 at 11 33 56


kentcdodds (approved):
Thanks!

PR #599, opened 2024-01-17
chore(notes): border animation

hakimLyon:
## Screenshots

https://github.com/epicweb-dev/epic-stack/assets/64654633/5dec58f3-4536-4650-8850-8a34c681f4b8


kentcdodds (changes requested):
This looks cool! But definitely not something that we want included by default. Would you be interested in making [an example](https://github.com/epicweb-dev/epic-stack/blob/main/docs/examples.md) instead?

PR #603, opened 2024-01-19
replace LoaderFunction with LoaderFunctionArgs in seo.md

JLarky:
This is not a huge change, but just to be consistent with the code, since `LoaderFunctionArgs` is used everywhere instead of `LoaderFunction`


kentcdodds (approved):
Thanks!

PR #607, opened 2024-01-23
Migrate to conform v1

edmundhung:
> This is a draft based on v1.0.0-rc.0 and require no action until v1 is released

Here you can find a summary of all the changes required to upgrade to Conform v1.

- All the conform helpers are renamed, e.g. `conform.input` -> `getInputProps`
- Both `parse` and `getFieldsetConstraint` are renamed to `parseWithZod` and `getZodConstraint`
- Setting up the form using the `getFormProps` helper instead of `form.props` which is no longer available
- Control the submission flow using `submission.status` and manipulate it using the new `submission.reply()` API
- Removed the status from the server response as it is now covered with the submission API and is accessible through `form.status`

### DIFF
--- a/app/components/forms.tsx
+++ b/app/components/forms.tsx
@@ -1,5 +1,5 @@
-import { useInputEvent } from '@conform-to/react'
-import React, { useId, useRef } from 'react'
+import { useInputControl } from '@conform-to/react'
+import React, { useId } from 'react'
 import { Checkbox, type CheckboxProps } from './ui/checkbox.tsx'
 import { Input } from './ui/input.tsx'
 import { Label } from './ui/label.tsx'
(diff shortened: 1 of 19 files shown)


kentcdodds:
Thank you for making this! It looks good to me. I'm seeing some improvements on top of what we have today and I don't see anything I don't really like!

I was expecting to see a form provider. I thought that was in the plans. Do we not need that afterall?

edmundhung:
> I was expecting to see a form provider. I thought that was in the plans. Do we not need that afterall?

I have thought of setting it up on the note editor form with the ImageChooser component. But end up just passing the metadata down to keep it simple.

The `<FormProvider />` is required only if any of your fields use the `useField()` or `useFormMetadata()` hook. You might consider using these hooks for two reasons:

1. Performance. These two hooks initiate a new subscription to the form context and will trigger a re-render only if the relevant metadata is changed. If you pass the metadata down, then the component will share the same subscription with the rest of the form and re-render together.
2. Sometimes you just couldn't pass the metadata down easily because of layout constraint. With `useField` or `useFormMetadata`, you can always pass a `name` or `formId` and retrieve the metadata anywhere down the tree.

kentcdodds:
Got it. Cool. Thank you!

kentcdodds:
Hey @edmundhung, thank you so much for this! It's extremely helpful that you put the work into migrating the Epic Stack for us!

kentcdodds:
I just noticed this bit doesn't appear to have been migrated, do you have advice on how to address this use case? https://github.com/epicweb-dev/epic-stack/blob/c9a17b3b164ca9fb2e1467ae7eb9aec570c68044/app/routes/_auth%2B/onboarding_.%24provider.tsx#L98-L108

edmundhung:
> I just noticed this bit doesn't appear to have been migrated, do you have advice on how to address this use case? https://github.com/epicweb-dev/epic-stack/blob/c9a17b3b164ca9fb2e1467ae7eb9aec570c68044/app/routes/_auth%2B/onboarding_.%24provider.tsx#L98-L108

I have thought about adding a `initialError` option to useForm but wanted to check if there are better alternative after v1 is out. Probably something to be addressed on v1.1

kentcdodds:
What do you recommend we do as a workaround until then?

taylrd:
Hey, thanks for all the hard work. Loving it! I found a typescript error in verify.tsx, line 257 on the actionData?.result within the useForm hook. "Property 'result' does not exist.

I thought I might try and fix it but I don't know how lol!

kentcdodds:
Hmmm... It seems to be passing type checking on CI https://github.com/epicweb-dev/epic-stack/actions/runs/7733353460/job/21085233897 🤔

taylrd:
Hmm, I wonder if it would be that Im running remix 2.6? I do not see anything else that it could be. I copied the entire file and dropped into my project and it gives the same error:

Property 'result' does not exist on type 'JsonifyObject<{ result: SubmissionResult<string[]>; }> | JsonifyObject<{ readonly status: "error"; readonly submission: Submission<{ type: "onboarding" | "reset-password" | "change-email" | "2fa"; code: string; target: string; redirectTo?: string | undefined; }, string[], { ...; }>; }>'.

I am not having an issue migrating any of the other files to Conform v1

taylrd:
Sorry Kent and Edmund, I should have caught this but I didn't notice handleChangeEmailVerification() was being imported and I had not migrated profile.change-email 🙄

Thanks again for showing us the way Kent and Edmund for the AWESOME form library!

kentcdodds:
Glad you worked it out!

PR #610, opened 2024-01-28
remove csrf

nichtsam:
CSRF protection technically isn't needed when using a combination of samesite-lax cookies and POST method endpoints. removed it to prevent confusion.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thank you for this!

Issue #609, opened 2024-01-26
Server-only module referenced by client


gregleeper:
I have borrowed the role based permissions model from the epic-stack. I am using it with the Vite Remix plugin. It notified me that the `#/utils/permissions.ts` file, which is client side, is referencing the `#/utils/auth.server.ts` file, which is server only code. 

```
[remix] 4:15:19 PM [vite] Internal server error: Server-only module referenced by client
[remix] 
[remix]     './auth.server.js' imported by 'app/utils/permissions.ts'
```

kentcdodds:
Ah yeah, that'll need to be refactored a bit to separate the server-only utils from the UI utils.

itsmegood:
Kent i looked through and found that permission.ts is only called a single place in the client https://github.com/epicweb-dev/epic-stack/blob/main/app/routes/users%2B/%24username_%2B/notes.%24noteId.tsx#L112

If you like we can remove either that and pass something is isOwner (boolean) down from loader and remove the code entirely and then convert permission.ts to fully server side.

kentcdodds:
The utility is useful for people to use as they build out their site, so we should split them up into a .server file for the server stuff and a regular file for the UI stuff

gregleeper:
For me, I moved `userHasPermission` and `userHasRole` to the `user.ts` utility as it made just as much sense there. I moved the rest of the `permissions.ts` file to the `permissions.server.ts` file.

kentcdodds:
This has been resolved. Thanks everyone!

PR #611, opened 2024-01-29
split permissions server and ui code

nichtsam:
solves #609

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (changes requested):
Thank you!

I like [the suggestion](https://github.com/epicweb-dev/epic-stack/issues/609#issuecomment-1914828380) @gregleeper made. Do you think you could move the `permissions.ts` contents to `user.ts`? I think that would make this even better.

nichtsam:
Definitely! I amended the commit to reflect the suggestion.

However I left `PermissionString` and `parsePermissionString` in `permissions.ts`.
I find this easier to reason about, otherwise it'd be `permissions.server.ts` importing these thing from `user.tsx`, and it somehow feels weird to me. Nevertheless, let me know if you would rather have it all moved into `user.tsx` and I would update again!

kentcdodds (changes requested):
I think I'm ok moving the parsing utility to user.ts along with the other permission stuff. I don't mind having the permissions.server.ts importing stuff from user.ts :)

Thanks!

kentcdodds (approved):
Lovely. Thank you!

PR #613, opened 2024-01-31
Use m1 runner to speed up deploy

xHomu:
> The new runner executes Actions workflows with a 3 vCPU, 7 GB RAM, and 14 GB of storage VM, which provides the latest Mac hardware Actions has to offer.
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### Typecheck 6s -> 4s
image

### Vitest 7s -> 4s
image

### Playwright Install 25s -> 5s

image

### Playwright Test 26s -> 21s
image

Honestly, the vanilla stack isn't big enough for it to make a notable difference, but it might for more built out apps? Are there any production builds we can check against?

(note: this deploy is a tad slower because of blank npm cache)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,7 +17,7 @@ permissions:
 jobs:
   lint:
     name: ⬣ ESLint
-    runs-on: ubuntu-22.04
+    runs-on: macos-14
     steps:
       - name: ⬇️ Checkout repo
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kiliman:
I agree that building on M1 would speed up the process. However, Epic Stack deploys to Fly using Docker on _bookwork-slim_, a Debian distro. The Ubuntu image used in the build is also Debian-based.

For CI, I think it's best if the build and testing use the same environment (or as close as possible) to production.

kentcdodds:
@kiliman said exactly what I would 👍

Thanks anyway @xHomu!

PR #616, opened 2024-02-09
move autoprefixer into devDependencies

nichtsam:
I noticed that `autoprefixer` is only used on build-time.
Tracked back to #168 and I believe it was accidentally moved to dependencies.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

PR #617, opened 2024-02-09
upgrade cachified

kentcdodds:
I've bumped into a little type issue. @Xiphe do you think you can help me figure out what's going on here?

### DIFF
--- a/app/utils/cache.server.ts
+++ b/app/utils/cache.server.ts
@@ -1,12 +1,14 @@
 import fs from 'fs'
 import {
 	cachified as baseCachified,
-	lruCacheAdapter,
 	verboseReporter,
 	mergeReporters,
 	type CacheEntry,
 	type Cache as CachifiedCache,
 	type CachifiedOptions,
+	type Cache,
+	totalTtl,
+	type Reporter,
 } from '@epic-web/cachified'
 import { remember } from '@epic-web/remember'
 import Database from 'better-sqlite3'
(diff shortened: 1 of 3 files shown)


app/utils/cache.server.ts
```diff
@@ -155,13 +169,14 @@ export async function searchCacheKeys(search: string, limit: number) {
 
 export async function cachified<Value>({
 	timings,
-	reporter = verboseReporter(),
+	reporter = verboseReporter<Value>(),
```
kentcdodds, line 172: The type error I'm getting on this line is:

> Type 'CreateReporter<Value>' is not assignable to type 'never'. ts(2322)

Xiphe: I left a `reporter?: never` on `CachifiedOptions` in order to warn people that their reporter will not function here. 

To keep the old signature you'd likely want to do this:

```ts
export async function cachified<Value>({
	timings,
	reporter = verboseReporter<Value>(),
	...options
}: Omit<CachifiedOptions<Value>, 'reporter'> & {
	timings?: Timings
	reporter?: Reporter<Value>
}): Promise<Value> {
	return baseCachified(
		options,
		mergeReporters(cachifiedTimingReporter(timings), reporter),
	)
}
```

Or adopt the new signature with:

```ts
export async function cachified<Value>({
	timings,
	...options
}: CachifiedOptions<Value> & {
	timings?: Timings
}, reporter: Reporter<Value> = verboseReporter<Value>()): Promise<Value> {
	return baseCachified(
		options,
		mergeReporters(cachifiedTimingReporter(timings), reporter),
	)
}
```

kentcdodds: I tried those, and with both of them I had the same error:

```
Type 'CreateReporter<Value>' is not assignable to type 'Reporter<Value>'.
  Types of parameters 'context' and 'event' are incompatible.
    Type 'CacheEvent<Value>' is not assignable to type 'Omit<Context<Value>, "report">'.
      Type 'GetFreshValueStartEvent' is missing the following properties from type 'Omit<Context<Value>, "report">': metadata, ttl, key, cache, and 6 more.ts(2322)
```

Xiphe: urgs. I'm gonna have another look

Xiphe: => https://github.com/epicweb-dev/epic-stack/pull/618

PR #618, opened 2024-02-12
Upgrade cachified

Xiphe:
updates to latest version of cachified and new reporter arg signature


kentcdodds (approved):
Ah, great! Thanks!

PR #624, opened 2024-02-22
upgrade to vite

kentcdodds:
Well this is great :)

https://remix.run/blog/remix-vite-stable


vite.config.ts
```diff
@@ -0,0 +1,36 @@
...
+						'.*',
+						'**/*.css',
+						'**/*.test.{js,jsx,ts,tsx}',
+						'**/__*.*',
+						// This is for server-side utilities you want to colocate next to
+						// your routes without making an additional directory.
+						// If you need a route that includes "server" or "client" in the
+						// filename, use the escape brackets like: my-route.[server].tsx
+						'**/*.server.*',
+						'**/*.client.*',
```
kentcdodds, line 30: This is probably the most significant *change*. It's the primary reason source code was changed. Remix with vite is more strict about your code that can end up in the client not using anything that's server-only so if you export anything other than the standard exports in a route that uses server-only code it'll error.

app/routes/_seo+/sitemap[.]xml.ts
```diff
@@ -1,10 +1,10 @@
 import { generateSitemap } from '@nasa-gcn/remix-seo'
-import { routes } from '@remix-run/dev/server-build'
-import { type LoaderFunctionArgs } from '@remix-run/node'
+import { type ServerBuild, type LoaderFunctionArgs } from '@remix-run/node'
 import { getDomainUrl } from '#app/utils/misc.tsx'
 
-export function loader({ request }: LoaderFunctionArgs) {
-	return generateSitemap(request, routes, {
+export async function loader({ request, context }: LoaderFunctionArgs) {
+	const serverBuild = (await context.serverBuild) as ServerBuild
```
Comment (kentcdodds) [inline app/routes/_seo+/sitemap[.]xml.ts:6]: This was another interesting change because importing `@remix-run/dev/server-build` is not allowed in vite.

server/index.ts
```diff
@@ -199,18 +194,27 @@ app.use((req, res, next) => {
...
-	MODE === 'development'
-		? (...args) => getRequestHandler(devBuild)(...args)
-		: getRequestHandler(build),
+	createRequestHandler({
+		getLoadContext: (_: any, res: any) => ({
+			cspNonce: res.locals.cspNonce,
+			serverBuild: getBuild(),
+		}),
+		mode: MODE,
+		// @sentry/remix needs to be updated to handle the function signature
```
kentcdodds, line 215: https://github.com/getsentry/sentry-javascript/issues/10724

PR #627, opened 2024-02-24
 Update Dockerfile to copy the correct build artifacts

hakimLyon:


other/Dockerfile
```diff
@@ -62,8 +62,8 @@ COPY --from=production-deps /myapp/node_modules /myapp/node_modules
 COPY --from=build /myapp/node_modules/.prisma /myapp/node_modules/.prisma
 
 COPY --from=build /myapp/server-build /myapp/server-build
-COPY --from=build /myapp/build /myapp/build
-COPY --from=build /myapp/public /myapp/public
+COPY --from=build /myapp/build/server /myapp/build/server
+COPY --from=build /myapp/build/client /myapp/build/client
```
kentcdodds, line 66: I think this should be sufficient.

```suggestion
COPY --from=build /myapp/build /myapp/build
```

kentcdodds (commented):
Thanks! I think this can be further simplified.

kentcdodds (approved):

PR #630, opened 2024-02-28
removed dev/main branches checks in condition during deployment; upda…

misiewiczmichal:
…ted comment with the same

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -140,10 +140,9 @@ jobs:
     name: 🚀 Deploy
     runs-on: ubuntu-22.04
     needs: [lint, typecheck, vitest, playwright]
-    # only build/deploy main branch on pushes
+    # only build/deploy branches on pushes
     if:
-      ${{ (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') &&
-      github.event_name == 'push' }}
+      ${{ github.event_name == 'push' }}
 
     steps:
       - name: ⬇️ Checkout repo


.github/workflows/deploy.yml
```diff
@@ -140,10 +140,9 @@ jobs:
     name: 🚀 Deploy
     runs-on: ubuntu-22.04
     needs: [lint, typecheck, vitest, playwright]
-    # only build/deploy main branch on pushes
+    # only build/deploy branches on pushes
     if:
-      ${{ (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev') &&
-      github.event_name == 'push' }}
+      ${{ github.event_name == 'push' }}
```
kentcdodds, line 145: Oh, actually, it just occurred to me that this may not be what we want. Could you verify that `github.event_name` is not `push` when running on a pull request?

kentcdodds (commented):
Perfect! Great work on your first PR. I just have one question.

.github/workflows/deploy.yml
misiewiczmichal: > Could you verify that github.event_name is not push when running on a pull request?

As I understand it, checking if it's a pull request will also use github.event_name so if
`github.event_name == 'push'`
then it will never be
`github.event_name == 'pull_request'`

Is this what you are asking or something else?

kentcdodds: Sounds good! Thanks!

Issue #635, opened 2024-03-02
Missing Hot Data reload??


natac13:
https://github.com/epicweb-dev/epic-stack/blob/430947ac60ba630d16209db86e0e860deb215045/server/index.ts#L199

I think this line has to be a function call which returns `ssrLoadModule` call. I noticed that without it being a function there is no hot data reload in a loader. HMR worked on the JSX but not in the loader. 

I changed this to a function in my app and things began to work again. 

I just wanted to share this info and to see if you were experiencing the same "no hot data reload"

kentcdodds:
Hmmm... I'm getting a refresh whenever I change my loader. Not HMR though. When I try to change that to a function I get an error because this is actually already in a function (the `getBuild` function) which is passed as-is to the `createRequestHandler` during development. So we are actually providing a function already.

Like I said, I'm getting a full-page refresh when I change my loader so that seems wrong too. I could be misunderstanding you though. Maybe it would help if you made a pull request?

PR #636, opened 2024-03-04
Demo for issue #635

natac13:
#635 

however, I have realized my error.
I was follow the epic stack in my remix app and was not ready for Sentry. Therefore the comment about it in `server/index.ts` made me this I could get rid of the NODE_ENV check. Therefore, I was left with just the `await getBuild()` call. However, because I was calling `getBuild` now in development the `viteDevServer.ssrLoadModule` part needed to be wrapped in a function like the remix/website example

To summarize. If you try this branch out and remove the wrapper function around `viteDevServer.ssrLoadModule` you will see that the app starts but updates inside a loader are not reflected. Which is where I thought to help

Please let me know if you would like more. Thank you for the Epic Stack and all your work

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -196,7 +196,7 @@ app.use((req, res, next) => {
 
 async function getBuild() {
 	const build = viteDevServer
-		? viteDevServer.ssrLoadModule('virtual:remix/server-build')
+		? () => viteDevServer.ssrLoadModule('virtual:remix/server-build')
 		: // @ts-ignore this should exist before running the server
 			// but it may not exist just yet.
 			await import('#build/server/index.js')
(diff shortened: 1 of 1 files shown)


natac13:
And just curious if your @ts-ignore is just for comment? Since it does not appear to be needed to avoid issues?

Screenshot 2024-03-04 at 1 37 13 PM

kentcdodds:
On the `@ts-ignore` bit, the `#build/server/index.js` file only exists after you run the build. But before you run the build, TypeScript will not be able to find it. That's why we have the `@ts-ignore`.

So if I understand you correctly, we don't need to make any changes because you mistakenly removed the `production` check when that actually has nothing to do with Sentry and actually is what supports HMR in the first place.

Thanks!

natac13:
Yep. I made the mistake

ajwtech:
I am not sure what just happened but either Github just changed something, or something related to this PR just broke the github extension in vscode. 
image
I use this to walk me through any PR's that are closed so I can to decide if I take them. I was just working on the Vite integration and then  this error started coming up. Any ideas? This is an odd one. But if anyone is curious this happens when I click on "load more" for a custom query "is closed"

PR #637, opened 2024-03-04
Update fly deploy workflow

devlinmorrow:
When I tried to set up and run the fly deploy workflow, I got errors for the steps below which said that they did not work with Node 20 and I needed to upgrade them. They seem to be working now with Node 20 with these newer versions.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -162,7 +162,7 @@ jobs:
           mv ./other/.dockerignore ./.dockerignore
 
       - name: 🎈 Setup Fly
-        uses: superfly/flyctl-actions/setup-flyctl@master
+        uses: superfly/flyctl-actions/setup-flyctl@1.5
 
       - name: 🚀 Deploy Staging
         if: ${{ github.ref == 'refs/heads/dev' }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -146,10 +146,10 @@ jobs:
 
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -162,7 +162,7 @@ jobs:
           mv ./other/.dockerignore ./.dockerignore
 
       - name: 🎈 Setup Fly
-        uses: superfly/flyctl-actions/setup-flyctl@v1.4
+        uses: superfly/flyctl-actions/setup-flyctl@master
```
devlinmorrow, line 165: for this one, I tried different syntax variations with the pinned version `v1.5` but for some reason it said that it doesn't exist, even though the repo for the action says it does exist. Maybe someone can get it working?

kentcdodds: I don't feel comfortable using `master`. There is a tag for [`v1.5`](https://github.com/superfly/flyctl-actions/releases/tag/1.5). Can you share a reproduction of this not working?

kentcdodds:
Thanks for this! Just want to figure out that `@master` thing.

.github/workflows/deploy.yml
devlinmorrow: image

devlinmorrow: Ah, good news, I tried removing the `v` from it, and it works. Dang, I thought I had already tried that, woopsie. Maybe give it a spin yourself before merging in to make sure though?

kentcdodds (approved):
Thank you!

ajwtech:
Hey guys, this was actually fixed last week. https://github.com/superfly/flyctl-actions/issues/54#event-11949716469 I switched back and it is working fine.

PR #639, opened 2024-03-06
Epic Stack example with Keystatic

simonswiss:
This adds the Keystatic example, as discussed (and screenshotted) in https://github.com/epicweb-dev/epic-stack/discussions/160#discussioncomment-8674164

The way the "skeleton" HTML scaffold for the Keystatic routes is being setup [might need some review/feedback](https://github.com/simonswiss/epic-stack-with-keystatic/commit/3959a0ab9dfed740c13eff7c0d00140b1f436d00) 😅


kentcdodds:
I think moving things out of `root.tsx` would be the right approach here. Have you tried nesting all the current routes under `routes/app_+` and putting most of what's in `root.tsx` into `routes/app_+/_layout.tsx`? That would allow you to have `routes/keystatic+/` and it could render the document without issue.

simonswiss:
I managed to move the root layout contents to `app/routes/_app+/_layout.tsx`.

Here's what the actual `root.tsx` file looks like now (I had to bring back the `nonce` to allow Keystatic to load properly):

https://github.com/simonswiss/epic-stack-with-keystatic/blob/main/app/root.tsx

And here's the common  layout in `app/routes/_app+/_layout.tsx`:

https://github.com/simonswiss/epic-stack-with-keystatic/blob/main/app/routes/_app%2B/_layout.tsx

kentcdodds (approved):
Very cool! Thanks for putting this example together. Probably lots of opportunity to show this off in some keystatic content 😉

PR #641, opened 2024-03-07
Update fonts documentation. 

ajwtech:
The move to Vite broke the implementation of custom fonts as described in the documentation. Added clarification, missing tailwinds import, as well as changes to be compatible with Vite.

## Test Plan
N/A just documentation

## Checklist

- [N/A] Tests updated
- [x] Docs updated


kentcdodds (approved):
Thank you!

Issue #647, opened 2024-03-12
ignorePatterns missing from autocompletion for eslint config


fredericrous:
the following type does not give autocompletion for some valid eslint configuration properties like ignorePatterns

https://github.com/epicweb-dev/epic-stack/blob/9741885c6fadc715ff502356164c6ec92782e1a2/.eslintrc.cjs#L5

this does

```js
/** @type {import('@types/eslint').Linter.Config} */ 
```

PR #648, opened 2024-03-12
eslintrc typings

fredericrous:
eslintrc was typed with BaseConfig instead of Config

this fixes https://github.com/epicweb-dev/epic-stack/issues/647


kentcdodds (approved):
Oh nice! Thanks!

Issue #651, opened 2024-03-12
sentry.io flood playwright tests output


fredericrous:
After configuring env variables SENTRY_DSN and co, I get the following warnings when I run the playwright tests

```console
  • POST https://xxxxxxx.ingest.us.sentry.io/api/xxxxxxx/envelope/?sentry_key=xxxxxxx&sentry_version=7&sentry_client=sentry.javascript.remix%2F7.106.0

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
[WebServer] [MSW] Warning: intercepted a request without a matching request handler:
```

the console gets flooded.

proposed solution: configure msw to blacklist sentry

PR #652, opened 2024-03-12
configure msw to blacklist sentry

fredericrous:
fix https://github.com/epicweb-dev/epic-stack/issues/651

### DIFF
--- a/tests/mocks/index.ts
+++ b/tests/mocks/index.ts
@@ -5,7 +5,17 @@ import { handlers as resendHandlers } from './resend.ts'
 
 export const server = setupServer(...resendHandlers, ...githubHandlers)
 
-server.listen({ onUnhandledRequest: 'warn' })
+server.listen({
+  onUnhandledRequest(request, print) {
+    // Do not print warnings on unhandled requests to Sentry.
+    if (request.url.includes('.sentry.io')) {
+      return
+    }
+
+    // Print the regular MSW unhandled request warning otherwise.
+    print.warning()
+  }
+})
 
 if (process.env.NODE_ENV !== 'test') {
 	console.info('🔶 Mock server installed')
--- a/tests/mocks/index.ts
+++ b/tests/mocks/index.ts
@@ -7,7 +7,10 @@ export const server = setupServer(...resendHandlers, ...githubHandlers)
 
 server.listen({
   onUnhandledRequest(request, print) {
-    // Do not print warnings on unhandled requests to Sentry.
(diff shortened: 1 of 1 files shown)


kentcdodds (changes requested):
I think I'd rather add a handler for these and use [passthrough](https://mswjs.io/docs/api/passthrough/).

fredericrous:
I would you achieve this with passthrough?

we cannot do something like this

```ts
const miscHandlers = process.env.REMIX_DEV_ORIGIN ? [
    http.post(`${process.env.REMIX_DEV_ORIGIN}ping`, passthrough),
    process.env.SENTRY_DSN && http.post(process.env.SENTRY_DSN, passthrough)
].filter(Boolean) : []
```

because the url we have in SENTRY_DNS is something like https://user@password.ingest.us.sentry.io/project-key

but the post request is actually made towards

```console
https://user.ingest.us.sentry.io/api/project-key
```

there is an extra /api

therefore I cannot think of an elegant solution with passthrough without getting hacky

the solution I proposed in this PR used to be in the doc if I remember correctly, at least it's the one I have been using for years. I actually did not bother looking in my old projects and googled it to get this peace of code from the author of mswjs https://github.com/mswjs/msw/discussions/1589#discussioncomment-5579717

kentcdodds:
Could we not add `http.get('https://user.ingest.us.sentry.io/api/project-key', passthrough)`?

fredericrous:
"user" here is unique for everyone, I could do this and it would only work for me and I'll be sharing a username that was private before (defined by env variable)

using glob pattern is not supported. the following won't work

```console
  http.post('https://*.ingest.us.sentry.io/api/', passthrough)
```

using parameters won't work in this situation neither because :userId is a cname

```console
  http.post('https://:userId.ingest.us.sentry.io/api/:projectId/envelope', passthrough)
```

There is actually a feature request to intercept urls with more granularity

https://github.com/mswjs/msw/issues/1804

kentcdodds:
Gotcha. In that case, could you add a comment above the code to reference that issue and then we can merge this :)

kentcdodds (approved):
Thanks!

jacobparis:
It looks like this commit didn't make it into main, possible got mulched by a later rebase?

https://github.com/fredericrous/epic-stack/blob/c423926f232dfb6d18ee67a72d1074f893ec7c9e/tests/mocks/index.ts

kentcdodds:
Hmm... What are we missing? Looks like it's in main right now: https://github.com/epicweb-dev/epic-stack/blob/78ef2528cf3d6ae284bd183c85c9a7270a90b933/tests/mocks/index.ts

PR #658, opened 2024-03-16
Do not render the errors list container if there are no errors

gnapse:
This container, when empty, creates extra space below the field for no reason.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/components/forms.tsx
+++ b/app/components/forms.tsx
@@ -50,9 +50,11 @@ export function Field({
 				aria-describedby={errorId}
 				{...inputProps}
 			/>
-			<div className="min-h-[32px] px-4 pb-3 pt-1">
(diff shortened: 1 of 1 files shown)


gnapse:
Hmmm, I just realized that this may be intentional, to have that space available for when the error appears it does not cause a jump in the UI.

But it is weird to have a component with this extra margin at the bottom. It gets in the way when you do not want to have that space below it. Anyway, I now suspect that this won't be accepted. Let me know and I can close it.

kentcdodds:
Yes, this is intentional. Jumpy UI is worse than extra space. But as usual everyone will be modifying this to their own taste and design anyway. Thanks!

jacobparis:
You can always add `-mb-8` to the component when you use it if in that case you know it will never have errors, which will take away the empty space

PR #660, opened 2024-03-17
Changes for vite-sentry plugin

devlinmorrow:
I've just gone through the process of setting up sentry for a fresh deploy of this project. I found I had to make the changes below, please feel free to disregard and correct me if I have just misunderstood how it's supposed to be done :) 

## Test Plan

- Fresh deploy and set up sentry

## Checklist

- [x] Docs updated


fly.toml
```diff
@@ -12,9 +12,6 @@ auto_rollback = true
 source = "data"
 destination = "/data"
 
-[deploy]
-release_command = "node ./other/sentry-create-release"
```
devlinmorrow, line 16: as far as I understand, this file has been superseded by use of the `sentryVitePlugin`, which does the release and sourcemaps upload and deletion for you

kentcdodds (changes requested):
This is great! Could you update everything to just use the plug-in? Then we don't need to worry about the docs for "if you use the plug-in"

other/Dockerfile
```diff
@@ -29,6 +29,14 @@ RUN npm prune --omit=dev
 # Build the app
 FROM base as build
 
+# vars needed for the build (in vite.config)
+#ARG COMMIT_SHA
+#ENV COMMIT_SHA=$COMMIT_SHA
+#ARG SENTRY_AUTH_TOKEN
```
johnpaulkiser, line 35: Do not pass secrets as build args in docker, this is bad practice.

We want to use a docker build secret instead.

We'd need to do something like this at the build step 👇
```
RUN --mount=type=secret,id=SENTRY_AUTH_TOKEN sh -c 'export SENTRY_AUTH_TOKEN=$(cat /run/secrets/SENTRY_AUTH_TOKEN) && npm run build'
```

johnpaulkiser (changes requested):

other/Dockerfile
johnpaulkiser: We can pass this secret in the github action 

```yaml
      - name: 🚀 Deploy Production
        if: ${{ github.ref == 'refs/heads/main' }}
        run: |
          flyctl deploy --remote-only \
           --build-arg COMMIT_SHA=${{ github.sha }} 
           --build-secret SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
```

devlinmorrow: thanks! done!

johnpaulkiser: beautiful! 🫶

johnpaulkiser (approved):

devlinmorrow:
Hi @kentcdodds , that last commit was just adding a commented-out line... so I think the playwright test must just be flakey, not an error introduced by this commit/PR ...

docs/monitoring.md
```diff
@@ -20,22 +20,26 @@ You can sign up for Sentry and create a Remix project from visiting
...
-organization slug (`SENTRY_ORG_SLUG`).
+organization slug (`SENTRY_ORG`), and the slug name for your project too (`SENTRY_PROJECT`).
 
-Now, set the secrets in Fly.io:
+Uncomment the sentry-vite plugin in `vite.config` and the build command in the 'build' section of the Dockerfile, which shows how to pass the `SENTRY_AUTH_TOKEN` (below) as a docker secret, so it is available to the vite config when `npm run build` is run (you will need to do the same for `SENTRY_ORG` and `SENTRY_PROJECT`). Note that these do not need to be added to the `env.server` env vars schema, as they are only used during the build and not the runtime.
 
-```sh
-fly secrets set SENTRY_DSN=<your_dsn> SENTRY_AUTH_TOKEN=<your_auth_token> SENTRY_ORG=<your_org_slug> SENTRY_PROJECT=javascript-remix
-```
+The plugin will create sentry releases for you and automatically associate commits during the vite build if you set up a GitHub integration with it. In this setup we have utilised a simple strategy for naming releases of using the commit sha, passed in as a build arg via the GitHub action workflow.
```
kentcdodds, line 44:
```suggestion
The plugin will create sentry releases for you and automatically associate commits during the vite build once you set the `SENTRY_AUTH_TOKEN` in [your GitHub action secrets](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions). In this setup we have utilized a simple strategy for naming releases of using the commit sha, passed in as a build arg via the GitHub action workflow.
```

kentcdodds (changes requested):
Thanks for the work on this! Just a few things. Primarily around running the `npm run format` script and also making this work without having to uncomment code and instead just having to set the `SENTRY_AUTH_TOKEN` in GitHub Actions secrets which effectively is what enables sentry.

other/Dockerfile
```diff
@@ -37,6 +41,9 @@ ADD prisma .
 RUN npx prisma generate
 
 ADD . .
+
+# Replace the run command when setting up sentry
+# RUN --mount=type=secret,id=SENTRY_AUTH_TOKEN sh -c 'export SENTRY_AUTH_TOKEN=$(cat /run/secrets/SENTRY_AUTH_TOKEN) && npm run build'
 RUN npm run build
```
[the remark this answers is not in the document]
devlinmorrow: I tried this and various other formulations, but it never liked it when I made setting the auth token as a separate command to the npm build so I just made the change to remove the `sh -c`

vite.config.ts
```diff
@@ -22,15 +27,35 @@ export default defineConfig({
...
+				project: process.env.SENTRY_PROJECT,
+				release: {
+					name: process.env.COMMIT_SHA,
+					setCommits: {
+						auto: true,
+					},
+				},
+				sourcemaps: {
+					filesToDeleteAfterUpload: await glob([
+						'./public/**/*.map',
```
kentcdodds, line 55: Are there any files that match this glob?

vite.config.ts
```diff
@@ -22,15 +27,35 @@ export default defineConfig({
...
+						// directory. If you need a route that includes "server" or
+						// "client" in the filename, use the escape brackets like:
+						// my-route.[server].tsx
 						'**/*.server.*',
 						'**/*.client.*',
 					],
 				})
 			},
 		}),
+		process.env.SENTRY_AUTH_TOKEN ??
```
kentcdodds, line 41: I'd prefer to use a ternary here:

```ts
process.env.SENTRY_AUTH_TOKEN ? sentryVitePlugin({ /* config */ }) : null
```

other/Dockerfile
```diff
@@ -37,7 +40,11 @@ ADD prisma .
 RUN npx prisma generate
 
 ADD . .
-RUN npm run build
+
+# Mount the secret and set it as an environment variable and run the build
+RUN --mount=type=secret,id=SENTRY_AUTH_TOKEN \
+    export SENTRY_AUTH_TOKEN=$(cat /run/secrets/SENTRY_AUTH_TOKEN) && \
+    npm run build
```
kentcdodds, line 47: This is better 👍

Can you confirm this works if you don't have a `SENTRY_AUTH_TOKEN` when you run the docker build?

kentcdodds (changes requested):
You're doing great work here. I'm excited to get this feature in! Just a couple small things.

vite.config.ts
fredericrous: ternary seems harder to read in a big config, I'm gonna have to look where's the `:` to finally find all that is declaring is a disappointing null

vite.config.ts
```diff
@@ -22,15 +27,35 @@ export default defineConfig({
...
+			sentryVitePlugin({
+				disable: MODE !== 'production',
+				authToken: process.env.SENTRY_AUTH_TOKEN,
+				org: process.env.SENTRY_ORG,
+				project: process.env.SENTRY_PROJECT,
+				release: {
+					name: process.env.COMMIT_SHA,
```
fredericrous, line 48: good job on setting the release name based on the commit sha. It will help people that are on a monorepo with pnpm. indeed when using a custom dockerfile with pnpm deploy, there is no ./git folder and sentry fails to automatically generate a release name

other/Dockerfile
fredericrous: cat command will fail if there is no secret but the exit code will be ignored because it is executed in a subprocess
image
also the error is written on stderr, it is not set on the variable
image

vite.config.ts
fredericrous: if I read correctly, this peace of code, when SENTRY_AUTH_TOKEN is defined, sentryVitePlugin won't be included. so yeah it seems like the contrary of what we want

vite.config.ts
```diff
@@ -22,15 +27,35 @@ export default defineConfig({
...
 						'**/__*.*',
-						// This is for server-side utilities you want to colocate next to
-						// your routes without making an additional directory.
-						// If you need a route that includes "server" or "client" in the
-						// filename, use the escape brackets like: my-route.[server].tsx
+						// This is for server-side utilities you want to colocate
+						// next to your routes without making an additional
+						// directory. If you need a route that includes "server" or
+						// "client" in the filename, use the escape brackets like:
+						// my-route.[server].tsx
```
fredericrous, line 34: are you using a prettier config different than the one in this repo? this text should not have changed

vite.config.ts
devlinmorrow: so, when I ssh into my own app project, I don't see that the `public` folder has any `.map`s in it. But I did notice that `.server-build` does, so I've replaced the dirs. What do you think ?

vite.config.ts
devlinmorrow: 👍

other/Dockerfile
devlinmorrow: Confirmed, yep, I removed the build secret from the `deploy.yml` and my app still built and deployed (but failed to send the source maps without the auth token, as you'd expect)

devlinmorrow: Confirmed, I removed the secret from github for my own project and deploy succeeded fine, obviously without sending source maps due to missing auth token

vite.config.ts
devlinmorrow: I have not touched the prettier config in this repo. When I run `npm run format`, there are about 10 different files that change, I'd guess from people who did not run format before merging changes. 
cc @kentcdodds 

This is what I see when I pull the epic-stack and run `npm run format` 

image

kentcdodds: Yeah, that does happen sometimes. I'm also confused by prettier changing the comments here (I've never seen it do that), but I don't mind 🤷‍♂️

kentcdodds (approved):
This is great! Thank you very much for working on this 😃

kentcdodds:
We're getting Prisma issues on the E2E tests that are likely unrelated to this PR so I'm going to merge anyway. I have no idea what's going on with Prisma recently.

package.json
```diff
@@ -62,7 +62,8 @@
     "@remix-run/react": "2.8.0",
     "@remix-run/server-runtime": "2.8.0",
     "@sentry/profiling-node": "^7.105.0",
-    "@sentry/remix": "^7.105.0",
+    "@sentry/remix": "^7.107.0",
+    "@sentry/vite-plugin": "^2.15.0",
```
fredericrous, line 66: noticing this now but @sentry/vite-plugin should be in dev deps!? @devlinmorrow

vite.config.ts
fredericrous: I think that occurs coz people forget to install prettier on github ui (me the first). I don't know what's the best solution..an auto commit in a PR seems very annoying. an autocommit on merge seems unsafe. a pr check seems overkill

package.json
kentcdodds: Yes, I believe we could move this to dev deps. PR welcome!

PR #661, opened 2024-03-17
increase num of workers for playwright

sys13:
makes playwright faster. Could also add it to `test:e2e:run`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -126,7 +126,7 @@ jobs:
         run: npm run build
 
       - name: 🎭 Playwright tests
-        run: npx playwright test
+        run: npx playwright test --workers=2
 
       - name: 📊 Upload report
         uses: actions/upload-artifact@v3


kentcdodds (commented):
I believe playwright selects the number of workers automatically though. Why is it better to specify like this?

fredericrous:
the cli says workers are already set to 50%

```sh
npx playwright test -h
```

image

what could speed up the tests is to use a wasm version of base32 converter in `@epic-web/totp` or if that does not exist and if it is possible: convert to another base that has a wasm package published on npm already

sys13:
When running it locally without specifying `--workers` it is using only 1 worker. I'm running on an m3 Mac. Is it using more workers for other folks locally? 

The change probably doesn't make sense in CI/CD. Looks like default cores for a GitHub runner is only [2 anyways for private repos](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

kentcdodds:
I'm not 100% certain what the heuristics are, but I've definitely seen Playwright choose up to 5 workers on my machine before. I'd prefer to defer to playwright's own heuristics though considering everyone's machines will be different. 

Thanks anyway!

PR #663, opened 2024-03-19
fix sentry cors issues

fredericrous:
sentry changed its url from *.ingest.sentry.io to *.ingest.[LOCATION].sentry.io

which results in cors errors

```
[Report Only] Refused to connect to '.......' because it violates the following Content Security Policy directive: "connect-src *.ingest.sentry.io 'self'".
```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -24,7 +24,7 @@ be done in the `server/index.ts` file.
 			directives: {
 				'connect-src': [
 					MODE === 'development' ? 'ws:' : null,
-					process.env.SENTRY_DSN ? '*.ingest.sentry.io' : null,
+					process.env.SENTRY_DSN ? '*.sentry.io' : null,
 					"'self'",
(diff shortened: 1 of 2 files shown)


kentcdodds (commented):
Could you link me to docs that talk about this? I have a feeling they wouldn't change this on a whim. I would even expect an email to customers about this to avoid CSP issues surprises like this.

fredericrous:
I think you are correct on the fact sentry didn't change server url. However for new customers or for customers that are non US, we get assigned a different address with the location in it. The documentation regarding this change seems very light. all info I could find were this doc page that says the ability to change location is only available by creating a new accounts. see https://docs.sentry.io/product/accounts/choose-your-data-center/

I invite you to create a new account and follow the instructions in the readme of epic stack or just wait for people here to confirm the issue/fix. No hurry at all

kentcdodds (approved):
Thank you!

PR #667, opened 2024-03-23
docs(routing): Point to vite.config.js instead of remix.config.js

machour:
remix.config.js went away with the Vite upgrade


kentcdodds (approved):
Thanks!

PR #668, opened 2024-03-24
Server code refactoring and minor docs updates

rustworthy:
## Test Plan

- Removed duplicated globals patching in dev more;
- sentry/remix can now handle `getBuild` [correctly](https://github.com/getsentry/sentry-javascript/compare/7.105.0...7.107.0#diff-73298ca0f16878e3602c16356e48fdd7d13fc660510f2475300138185eb0d712R58-R86)
- Updated docs on monitoring.

## Checklist

- [ ] Tests updated N/A
- [x] Docs updated

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -143,7 +143,7 @@ app.use(
 // rate limiting because playwright tests are very fast and we don't want to
 // have to wait for the rate limit to reset between tests.
 const maxMultiple =
-	MODE !== 'production' || process.env.PLAYWRIGHT_TEST_BASE_URL ? 10_000 : 1
+	!IS_PROD || process.env.PLAYWRIGHT_TEST_BASE_URL ? 10_000 : 1
 const rateLimitDefault = {
 	windowMs: 60 * 1000,
 	max: 1000 * maxMultiple,
--- a/server/index.ts
+++ b/server/index.ts
@@ -218,35 +218,31 @@ app.all(
 )
 
 const desiredPort = Number(process.env.PORT || 3000)
-const portToUse = await getPort({
+const portToUser = await getPort({
 	port: portNumbers(desiredPort, desiredPort + 100),
 })
 
-const server = app.listen(portToUse, () => {
+const server = app.listen(portToUser, () => {
 	const addy = server.address()
-	const portUsed =
-		desiredPort === portToUse
-			? desiredPort
-			: addy && typeof addy === 'object'
-				? addy.port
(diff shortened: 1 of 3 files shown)


server/index.ts
```diff
@@ -212,41 +213,36 @@ app.all(
...
-const portToUse = await getPort({
+const portToUser = await getPort({
 	port: portNumbers(desiredPort, desiredPort + 100),
 })
 
-const server = app.listen(portToUse, () => {
+const server = app.listen(portToUser, () => {
 	const addy = server.address()
-	const portUsed =
-		desiredPort === portToUse
```
rustworthy, line 228: A tine [chance](https://github.com/sindresorhus/get-port?tab=readme-ov-file#beware), but still it makes sense to only trust the OS's final decision ;)

server/index.ts
```diff
@@ -108,6 +107,8 @@ app.use((_, res, next) => {
 
 app.use(
 	helmet({
+		// https://github.com/helmetjs/helmet?tab=readme-ov-file#x-powered-by
```
kentcdodds, line 110: We don't need to include links for every single thing here. People can look it up if they want.

```suggestion
```

server/index.ts
```diff
@@ -212,41 +213,36 @@ app.all(
...
 		mode: MODE,
-		// @sentry/remix needs to be updated to handle the function signature
-		build: MODE === 'production' ? await getBuild() : getBuild,
+		build: getBuild,
 	}),
 )
 
 const desiredPort = Number(process.env.PORT || 3000)
-const portToUse = await getPort({
+const portToUser = await getPort({
```
kentcdodds, line 221: Looks like a typo?

```suggestion
const portToUse = await getPort({
```

server/index.ts
```diff
@@ -212,41 +213,36 @@ app.all(
...
 	port: portNumbers(desiredPort, desiredPort + 100),
 })
 
-const server = app.listen(portToUse, () => {
+const server = app.listen(portToUser, () => {
```
kentcdodds, line 225: Typo?
```suggestion
const server = app.listen(portToUse, () => {
```

kentcdodds (changes requested):
Thanks! Just a few suggestions.

server/index.ts
rustworthy: It is

kentcdodds (approved):
Thanks!

PR #673, opened 2024-03-26
Port unavailable handling

rustworthy:
On staging, production or any other deployed environment, we (dockerd/proxy/balancer) need to know the port the service will be available on ahead of time. So if it is `3000` and this port is taken on the machine, we should not be trying to re-assign. This re-assigning only makes sense during  the development.

The callback passed to Express `app.listen` will not accept an error that we could immediately logged and so we would be able  to terminate the process.

However we can utilize the `portToUse` returned from `getPort` utillity, and if it is not the same as the desired one AND the environment is not `"development"`, we should immediately exit with non zero.

To test out, launch a session and hit `npm run dev`, then launch another one and issue the same command, you will see we've got 3001 assigned as expected. Then run `npm run build` and repeat the same test but this time with `npm run start`. You should see the process terminated with the corresponding message logged out.

I apologize for [this](https://github.com/epicweb-dev/epic-stack/pull/668#discussion_r1536848461) - I mistakenly thought that if we are not handling the port-already-in-use error in the `app.listen` callback, that means the app asked the OS for a new one for us. That's not the case. And this, anyhow, should - I guess - be only dev env behavior.

## Checklist

- [ ] Tests updated N/A
- [ ] Docs updated N/A

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -16,6 +16,7 @@ installGlobals()
 
 const MODE = process.env.NODE_ENV ?? 'development'
 const IS_PROD = MODE === 'production'
+const IS_DEV = MODE === "development"
 
 const createRequestHandler = IS_PROD
 	? Sentry.wrapExpressCreateRequestHandler(_createRequestHandler)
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
This change makes sense. Thanks!

Issue #671, opened 2024-03-25
Create an account button shows error but should not


fredericrous:
**Reproduction workflow**

1. setup the project without database seed:
    - clone the project
    - cd to the cloned folder
    - cp .env.example .env
    - npm install
    - setup without seed: npx prisma generate && npx prisma migrate deploy
    - npm run dev

3. then, connect with github:
    - click the login with github button
    - get automatically redirected to `/onboarding/github?redirectTo=%2Flogin`
    - notice the status on the button tells the user there is an error but there is none

<img width="1095" alt="image" src="https://github.com/epicweb-dev/epic-stack/assets/702227/c1f622f3-752f-4182-a858-8ad74b3c1a2f">

kiliman:
Seed data is required for the Epic Stack to function correctly. It creates the roles and permissions needed for account creation.

I'm unsure why you would not want to run the setup process.

kentcdodds:
The seed data is no longer required for the Epic Stack to function. We seed roles and permissions in the migrations.

Help is wanted to investigate what's going on here.

kentcdodds:
Thanks @rustworthy!!

Issue #657, opened 2024-03-15
[documentation] missing for github auth


fredericrous:
It'd be nice to get doc on github auth setup in https://github.com/epicweb-dev/epic-stack/blob/main/docs/deployment.md

I'm thinking of a link to github oauth settings page, the instructions to enter like the url and the callback

a link to the token and the authorization to select

kentcdodds:
Good idea! We do cover it here: https://auth.epicweb.dev/17

But having some docs specifically about setting up the GitHub side would be good

fredericrous:
is `read:user` the correct authorizations for the token expected by GITHUB_TOKEN?

kentcdodds:
I just follow the instructions here: https://github.com/sergiodxa/remix-auth-github

kentcdodds:
Thanks @rustworthy!

PR #675, opened 2024-03-26
Fix StatusButton status when creating account. Add docs on GitHub OAuth app set up

rustworthy:
Addressing #671 and #657.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/tests/mocks/github.ts
+++ b/tests/mocks/github.ts
@@ -118,7 +118,9 @@ async function getUser(request: Request) {
 }
 
 const passthroughGitHub =
-	!process.env.GITHUB_CLIENT_ID.startsWith('MOCK_') && !process.env.TESTING
+	!process.env.GITHUB_CLIENT_ID.startsWith('MOCK_') &&
+	process.env.NODE_ENV !== 'test'
+
 export const handlers: Array<HttpHandler> = [
 	http.post(
 		'https://github.com/login/oauth/access_token',
--- a/app/routes/_auth+/onboarding_.$provider.tsx
+++ b/app/routes/_auth+/onboarding_.$provider.tsx
@@ -92,16 +92,15 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
 	const prefilledProfile = verifySession.get(prefilledProfileKey)
 
 	const formError = authSession.get(authenticator.sessionErrorKey)
+	const hasError = typeof formError === 'string'
 
 	return json({
 		email,
 		status: 'idle',
 		submission: {
-			status: 'error',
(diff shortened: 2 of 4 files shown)


tests/mocks/github.ts
```diff
@@ -118,7 +118,9 @@ async function getUser(request: Request) {
 }
 
 const passthroughGitHub =
-	!process.env.GITHUB_CLIENT_ID.startsWith('MOCK_') && !process.env.TESTING
+	!process.env.GITHUB_CLIENT_ID.startsWith('MOCK_') &&
+	process.env.NODE_ENV !== 'test'
```
rustworthy, line 122: This looks like a left-over or a typo and not described by the env schema.  It led to the tests failing since we are bypassing the msw once real creds for GitHub OAuth app are set up.

rustworthy:
As for the `NODE_ENV` in general, there are [recommendations](https://blog.platformatic.dev/handling-environment-variables-in-nodejs#heading-set-nodeenvproduction-for-all-environments) to avoid relying on it, since 3rd party libs can have various behavior (Nestjs, Express and others), depending on this var. 

We could instead introduce our own dedicated variable e.g. `CURRENT_ENVIRONMENT` ("development" | "test" | "production") and set the `NODE_ENV` to "production" due to security concerns.

A report by Matteo Colina with a demo can be found [here](https://youtu.be/HMM7GJC5E2o?si=OGA8BbkLhJaxEfvT)

Wdyt?

kentcdodds (approved):
Oh nice, thank you! This is extremely helpful!

tests/mocks/github.ts
ryanwoodcox: Thanks for fixing this! Awesome to see someone actually going through and testing what is in this offering!

I posted in the epic-stack-help on 01/09/2024 and heard nothing.

https://discord.com/channels/715220730605731931/1194257050394443816/1194257050394443816
```Has anyone tried running npm run test after setting a valid GITHUB_CLIENT_ID in your .env file just to see a bunch of auth.$provider.callback.test.ts tests fail?

It seems const passthroughGitHub = !process.env.GITHUB_CLIENT_ID.startsWith('MOCK_') && !process.env.TESTING isn't quite right.

If you step through a debugging session you'll see process.env.TESTING is undefined.

Do we know if this check should be referencing a value that is actually defined such as:
process.env_MODE !== 'test' or 
process.env_NODE_ENV !== 'test' or
!process.env_TEST or
!process.env_VITEST

Thanks!```

PR #679, opened 2024-03-30
add control over indexability

nichtsam:
We don't over duplicate content from different environment to all be indexed by search engines, \
this PR adds control over indexability.

For seo safety reason, I decided it's better to allow indexing by default.

`<meta name="robots" content="noindex, nofollow" />` prevents HTML from being indexed.
`X-Robots-Tag` prevents all resource from being indexed, this is mainly for non-HTML resources, such as PDF files, video files, or image files.

## Test Plan

I don't see a test that brings much confidence here.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,3 +13,7 @@ SENTRY_DSN="your-dsn"
 GITHUB_CLIENT_ID="MOCK_GITHUB_CLIENT_ID"
 GITHUB_CLIENT_SECRET="MOCK_GITHUB_CLIENT_SECRET"
 GITHUB_TOKEN="MOCK_GITHUB_TOKEN"
+
+# set this to false to prevent search engines from indexing the website
+# default to allow indexing for seo safety
+ALLOW_INDEXING="true"
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -181,11 +181,13 @@ function Document({
 	nonce,
 	theme = 'light',
 	env = {},
+	allowIndexing = true,
 }: {
 	children: React.ReactNode
 	nonce: string
 	theme?: Theme
 	env?: Record<string, string>
+	allowIndexing?: boolean
 }) {
 	return (
 		<html lang="en" className={`${theme} h-full overflow-x-hidden`}>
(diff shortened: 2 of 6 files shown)


app/root.tsx
```diff
@@ -194,6 +196,7 @@ function Document({
 				<Meta />
 				<meta charSet="utf-8" />
 				<meta name="viewport" content="width=device-width,initial-scale=1" />
+				{disallowIndexing && <meta name="robots" content="noindex, nofollow" />}
```
kentcdodds, line 199:
```suggestion
				{allowIndexing ? null : <meta name="robots" content="noindex, nofollow" />}
```

.env.example
```diff
@@ -13,3 +13,7 @@ SENTRY_DSN="your-dsn"
 GITHUB_CLIENT_ID="MOCK_GITHUB_CLIENT_ID"
 GITHUB_CLIENT_SECRET="MOCK_GITHUB_CLIENT_SECRET"
 GITHUB_TOKEN="MOCK_GITHUB_TOKEN"
+
+# set this to true to prevent search engines from indexing the website
+# default to allow indexing for seo safety
+DISALLOW_INDEXING="false"
```
kentcdodds, line 19: I generally prefer variables not use negation. So this should instead be "ALLOW_INDEXING" which defaults to `true`

app/root.tsx
```diff
@@ -181,11 +181,13 @@ function Document({
 	nonce,
 	theme = 'light',
 	env = {},
+	disallowIndexing,
```
kentcdodds, line 184: This should be:

```suggestion
	allowIndexing = true,
```

kentcdodds (changes requested):
Thanks! This looks pretty good. Just a simple suggestion.

nichtsam:
I use `process.env.ALLOW_INDEXING !== 'false'` to get the boolean, so only explicitly setting it to false will disallow the indexing.
And I realized `.default()` isn't really safe in this case, because it would declare type as `"true" | "false"` while it's actually `"true" | "false" | undefined`, and since it's default to true, it's important to be aware of the `undefined` case.

kentcdodds (approved):
Super. Thank you!

PR #680, opened 2024-03-30
Test for onboarding after GitHub OAuth

rustworthy:
This is a chaser to #675

Adding e2e  test for GitHub OAuth onboarding and, since we are here, verifying the status button is idle when they first get onto the onboarding page.

## Checklist

- [x] Tests updated
- [ ] Docs updated


app/components/ui/status-button.tsx
```diff
@@ -25,17 +25,26 @@ export const StatusButton = React.forwardRef<
 	})
 	const companion = {
 		pending: delayedPending ? (
-			<div className="inline-flex h-6 w-6 items-center justify-center">
+			<div
+				role="status"
```
kentcdodds, line 29: Good idea on this. But we need to give these an accessible name so screen readers can announce what the current status is. I think we should probably add a `title` prop to `Icon` and if it's specified, then we add a `<title>` to the `<svg>` that's rendered (next to the `<use>`).

As an additional benefit, way you can select it by it's accessible name as well as its role.

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +137,139 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+
+	// button currently in 'idle' (neutral) state and so has got no companion
+	const createAccountButton = page.getByRole('button', {
+		name: /create an account/i,
+	})
+	expect(createAccountButton.getByRole('status')).not.toBeVisible()
+
+	// attempt 1
+	// Username is a string of alphanums and underscores from 3 to 20 chars
+	// long, that (NB!) will be lowercased upon sumbission. See: app/utils/user-validation.ts:
```
kentcdodds, line 187: What does "NB!" mean?

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +137,139 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+	const user = await prisma.user.findUnique({
+		where: { email: ghUser?.primaryEmail.toLowerCase() },
+	})
+	if (user) await prisma.user.delete({ where: { id: user.id } })
+
+	// we remove the connection if any to keep the test clean
+	const conn = await prisma.connection.findFirst({
+		where: { providerName: 'github', userId: user?.id },
+	})
+	if (conn) await prisma.connection.delete({ where: { id: conn.id } })
```
kentcdodds, line 159: Rather than finding an existing user and deleting them, I would prefer to restructure things so we create a brand new github user for this test.

kentcdodds (changes requested):
Thanks for this! I've looked over some of it and have a bit of feedback. Thanks!

tests/e2e/onboarding.test.ts
rustworthy: Means ["pay attention!"](https://en.wikipedia.org/wiki/Nota_bene), but those sneakers make made it confusing :smile:

tests/e2e/onboarding.test.ts
rustworthy: That was my original intention: each test should be isolated and have its own prep and clean up steps that will not impact the other tests, that's true. But I am now struggling a bit with the [code](https://github.com/epicweb-dev/epic-stack/blob/main/app/utils/providers/github.server.ts#L87) being a predefined value in  the `handleMockAction` method, and we're going to look up a user from our mocked github users database using this value. I am looking into it

rustworthy: All right, playwright gives us API to intercept requests and patch them, so used this approach. This gives us proper isolation. 

Just for the sake of discussion. The other option would be to generate (rather than have it hard coded) a `code` in the `mockHandleAction` method, insert a user with this code into the mocked github users database, and then redirect with this code in query string.

app/components/ui/status-button.tsx
rustworthy: How beautiful is that, the browser will also show a tooltip on hover

rustworthy:
> Thanks for this! I've looked over some of it and have a bit of feedback. Thanks!

This is totally my pleasure! Please have a look at it once again

playwright.config.ts
```diff
@@ -36,6 +36,7 @@ export default defineConfig({
 		stderr: 'pipe',
 		env: {
 			PORT,
+			NODE_ENV: 'test',
```
rustworthy, line 39: when running on CI, we `cp` the `.env.example` to `.env` and so we've got values staring with `MOCK_` which turns   on the [`shouldMock` ](https://github.com/epicweb-dev/epic-stack/blob/main/app/utils/providers/github.server.ts#L22) flag.

But when running e2e tests locally and having `.env` with real values for GitHub OAuth app, we need something else to indicate we want mock actions. Hence adding this env var

app/utils/providers/github.server.ts
```diff
@@ -19,7 +20,9 @@ const GitHubUserParseResult = z
 		}),
 	)
 
-const shouldMock = process.env.GITHUB_CLIENT_ID?.startsWith('MOCK_')
+const shouldMock =
+	process.env.GITHUB_CLIENT_ID?.startsWith('MOCK_') ||
+	process.env.NODE_ENV === 'test'
```
rustworthy, line 25: see [comment](https://github.com/epicweb-dev/epic-stack/pull/680/files#r1545761063)

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
 	await expect(page).toHaveURL(`/onboarding`)
 })
 
+test('onboarding with GitHub OAuth', async ({ page }) => {
+	const code = `${MOCK_CODE_GITHUB}_${crypto.randomUUID()}_PLAYWRIGHT_TEST`
```
kettanaito, line 144: If the intention is to create a random `code` for each test run, may I suggest [`testInfo.testId`](https://playwright.dev/docs/api/class-testinfo#test-info-test-id)? 

- Playwright already generates that;
- We get free collocation between the code and the test that generated it.

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+	const createAccountButton = page.getByRole('button', {
+		name: /create an account/i,
+	})
+	expect(createAccountButton.getByRole('status')).not.toBeVisible()
+	expect(createAccountButton.getByText('cross')).not.toBeAttached()
+
+	// attempt 1
+	// Username is a string of alphanums and underscores from 3 to 20 chars
+	// long, that will be lowercased upon sumbission. See: app/utils/user-validation.ts:
+	await usernameInput.fill('U$er_name') // $ is invalid char
```
kettanaito, line 181: I'd suggest we split the happy path and validation tests into separate test cases. 

- completes the github onboarding given valid user credentials
- renders a validation error given invalid username
- etc...

This way test failures will let us know more about the problematic area of the code.

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+
+	// log out
+	await page
+		.getByRole('link', { name: newlyCreatedUser.name as string })
+		.click()
+	await page.getByRole('menuitem', { name: /logout/i }).click()
+	await expect(page).toHaveURL(`/`)
+
+	// delete newly created connection, but keep the user:
+	const newlyCreatedConn = await prisma.connection.findFirst({
```
kettanaito, line 243: Nit: `const newConnection`?

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
 	await expect(page).toHaveURL(`/onboarding`)
 })
 
+test('onboarding with GitHub OAuth', async ({ page }) => {
+	const code = `${MOCK_CODE_GITHUB}_${crypto.randomUUID()}_PLAYWRIGHT_TEST`
+
+	await page.route(/\/auth\/github(?!\/callback)/, async (route, request) => {
```
kettanaito, line 146: It's perfectly fine as it is but since Epic Stack uses MSW already, we may take advantage of the shared network description between integration and E2E tests. 

There are a few request handlers defined for GitHub auth already:
https://github.com/epicweb-dev/epic-stack/blob/800022dae82530dd48ef0ba35abe666eac1c619e/tests/mocks/github.ts#L124-L125

In the future, we may refactor this and replace `page.route()` with the [browser-side integration](https://mswjs.io/docs/integrations/browser) of MSW to intercept this request and produce whichever response is needed (e.g. the one including a proper `code`).

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+
+	await expect(page).toHaveURL(/\/onboarding\/github/)
+	expect(
+		page.getByText(new RegExp(`welcome aboard ${ghUser.primaryEmail}`, 'i')),
+	).toBeVisible()
+
+	// fields are pre-populated for the user
+	const usernameInput = page.getByRole('textbox', { name: /username/i })
+	expect(usernameInput).toHaveValue(
+		// @ts-ignore
```
kettanaito, line 164: Why is this failing for TypeScript? 👀 Let's look deeper into this and resolve it so we don't have to use `@ts-ignore`, it's a bad practice.

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+		),
+	).toBeVisible()
+	await expect(page).toHaveURL(/\/onboarding\/github/)
+
+	// attempt 2
+	await usernameInput.fill('')
+	await createAccountButton.click()
+	await expect(page.getByText(/username is required/i)).toBeVisible()
+	// the next assertion will fail...
+	// await expect(createAccountButton.getByText('cross')).toBeAttached()
```
kettanaito, line 204: Is this commented assertion left here intentionally?

tests/e2e/onboarding.test.ts
```diff
@@ -131,6 +140,134 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
...
+	// await expect(createAccountButton.getByText('cross')).not.toBeAttached()
+	await expect(page).toHaveURL(/signup/i) // home page
+
+	// we are still on the 'signup' route since that
+	// was the referrer and no 'redirectTo' has been specified
+	await expect(page).toHaveURL('/signup')
+	await expect(page.getByText(/thanks for signing up/i)).toBeVisible()
+
+	// internally, a user is being created:
+	const newlyCreatedUser = await prisma.user.findUniqueOrThrow({
```
kettanaito, line 231: Is this a part of the testing setup (necessary for the next steps to pass) or an expected action in Prisma if the login flow reaches a certain state? 

Trying to figure out what this should do in relation to the test.

rustworthy: This is to show a user in being created in out db, and I see what you mean, we never verified this user was not there already. Added a check

tests/e2e/onboarding.test.ts
rustworthy: not at all, removed

tests/e2e/onboarding.test.ts
rustworthy: I only when for `ignore` directive because we are in tests and so are safely proving that things are correct _at runtime_. But I totally see your point and will now iterate over it once again.

rustworthy: Oh that was prosaic one: `toHaveValue` exprects a string or RegEx, while the return val of `normalizeUsername` is string or undefined. I did not give it enough info, like if string comes in, string definitely comes out and you should not be yelling at me :) Anyhow, I hated the fact that `normalizeUsername` can receive undefined, so I adjusted it to work with strings only.

tests/e2e/onboarding.test.ts
rustworthy: It's just so tempting to use such a clean API :)

tests/e2e/onboarding.test.ts
rustworthy: I initially wanted the code to resemble the one we see, but I know realize that binding it ti the test instance makes much more sense. updated

tests/e2e/onboarding.test.ts
rustworthy: I imagined this test to be a real life user journey, hence the layout and steps. But indeed, there are a couple of scenarios mixed up in here: happy registration path, happy account linking path, and struggling with the form path :sweat_smile: 

Please  have a lot at the decomposed version.

Also, I guess that's a nit, but we currently have inconsistent naming of tests (description) in e2e tests. The docs will also have varying examples, like [`test('does something')`](https://playwright.dev/docs/api/class-pageassertions) vs [`test('basic test')`](https://playwright.dev/docs/api/class-testinfo) . We also have both naming approaches in our unit [tests](https://github.com/epicweb-dev/epic-stack/blob/main/app/routes/_auth%2B/auth.%24provider.callback.test.ts#L140). I guess this is a matter of convention on the project.

kettanaito: I think splitting them by different behaviors would make the most sense. 

When it comes to test names, I highly suggest [these tips](https://www.epicweb.dev/talks/how-to-write-better-test-names). I find them immensely useful when naming test cases (and also arranging test suites).

tests/e2e/onboarding.test.ts
kettanaito: Yeah, it's perfectly fine!

tests/e2e/onboarding.test.ts
kettanaito: > like if string comes in, string definitely comes out and you should not be yelling at me

This sounds like a perfect use case for [generic types](https://kettanaito.com/blog/making-sense-of-typescript-generics)! Let's describe precisely this intention in types and get the correct type validation here without having to resort to ignoring the typecheck.

tests/e2e/onboarding.test.ts
kettanaito: Sorry, I'm still a bit confused. Let me phrase my question better:

- Is this database call a result of the user's interaction? Or,
- Is this database call required in order for future user interactions to even work correctly? 

Based on this, we need to approach this code differently.

rustworthy: there are two calls: the first one demonstrates the user is not in our system, the other one (which we are running after they signup)  - shows they are in db. And yes - later on (at the clean up stage) we re-use the ref to the db user

tests/e2e/onboarding.test.ts
rustworthy: True, but we do not resort to `ignore` any longer. Again, the `normalizeUsername` works with strings and so compiler can now see we are calling  `toLowerCase` on a string and returning the result

tests/e2e/onboarding.test.ts
rustworthy: tq, those are useful tips, I also like `prompts to confirm deletion` better than `account deletion` :)

app/components/ui/status-button.tsx
```diff
@@ -25,18 +25,31 @@ export const StatusButton = React.forwardRef<
...
 			</div>
 		) : null,
 		success: (
-			<div className="inline-flex h-6 w-6 items-center justify-center">
-				<Icon name="check" />
+			<div
+				role="status"
+				className="inline-flex h-6 w-6 items-center justify-center"
+			>
+				<Icon name="check" title="check mark" />
```
kentcdodds, line 40:
```suggestion
				<Icon name="check" title="success" />
```

app/components/ui/status-button.tsx
```diff
@@ -25,18 +25,31 @@ export const StatusButton = React.forwardRef<
...
-			<div className="inline-flex h-6 w-6 items-center justify-center rounded-full bg-destructive">
-				<Icon name="cross-1" className="text-destructive-foreground" />
+			<div
+				role="status"
+				className="inline-flex h-6 w-6 items-center justify-center rounded-full bg-destructive"
+			>
+				<Icon
+					name="cross-1"
+					className="text-destructive-foreground"
+					title="cross"
```
kentcdodds, line 51:
```suggestion
					title="error"
```

rustworthy:
> This is so rad. Thank you for working on it! I just have a couple things!

The pleasure is all mine! I thank you back, and also @kettanaito and all the members of the epic community for the great conference. I watched the first half (your morning - my evening) live and the record of the second half the next day.

As for this [thread](https://github.com/epicweb-dev/epic-stack/pull/680#discussion_r1566448593), I did spend some time on it after Artem's review, and did not find a way out. Besides, I found myself duplicating a lot of logic from the app (stealing btw some work from the mocked GitHub API) and so being dependent on the implementation details. Of note here is that we _are_ using the `msw` layer for the GitHub API which we _allow_ to hit in the variant I am suggesting. We are using the `Playwright` test runner API to pass through the request to GitHub (and we in this test do not know this is mocked GitHub) only patching the request headers. 

I also see a separation of concerns: we  are mocking 3rd party services with `msw` and only _patching_ request to our own API. My understanding is that this is a nice combo.  At the same time, I am _not_ excluding that there is a more elegant solution that I am missing :sweat_smile:

kentcdodds (changes requested):
Thanks for this! I think we should probably use [a fixture](https://playwright.dev/docs/test-fixtures) for this to reduce repetition and improve reliable cleanup. Here's the bit in epic web about this: https://testing.epicweb.dev/03

We have a fixture for the login state. It'll be similar to that.

I appreciate your work on this! I'm excited to get this merged!

rustworthy:
> Thanks for this! I think we should probably use [a fixture](https://playwright.dev/docs/test-fixtures) for this to reduce repetition and improve reliable cleanup. Here's the bit in epic web about this: https://testing.epicweb.dev/03
> 
> We have a fixture for the login state. It'll be similar to that.
> 
> I appreciate your work on this! I'm excited to get this merged!

The pleasure is all mine! Please check it out once again

kentcdodds (approved):
I love it. Thanks!

PR #683, opened 2024-04-01
Add Stripe example

saurabhp75:
## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Fantastic!

fredericrous:
imo this example needs a rework

- it gives a bad user experience because of the multi redirect here https://github.com/saurabhp75/epic-stripe/blob/baffb612ced3d53f53d17681baf31e62a2387b34/app/routes/account.tsx#L30

- I don't get the value of wrapping the stripe calls into endpoints https://github.com/saurabhp75/epic-stripe/blob/baffb612ced3d53f53d17681baf31e62a2387b34/app/services/stripe/api/create-customer.ts

- finally the duplication of data between prisma and stripe for the pricing plans makes thing complicated and is not what strip puts in its examples

saurabhp75:
I will fix first two items. 

As for third item, there are few points to consider :

- The app needs to keep some data locally (which will be synced using webhook). This is essential for better UX and avoid network latency. IMHO this is a pretty standard approach when integrating with payments platform.

- The data to keep in Prisma depends very much on the type of products being sold like one time purchases, subscriptions or the one with licences or usage base plans. See item below for the use cases this example handles.

- This example provides 3 subscription based products with option of monthly or yearly payments. Each subscription variant can be configured with a limit on number of units user can consume.

- Do let me know which Stripe example you are looking at for DB schema.

saurabhp75:
README updated based on the discussion.

PR #687, opened 2024-04-03
Update prisma to v5.12.0

rajeshg:
## Test Plan

Updated prisma to v5.12.  Previous commit fa43eeec0910630727c81bd546a251f477c4016f  is missing the changes from package.json. 
Also I updated the seed.ts script, to make use of the newly added createMany method in prisma for sqlite. See [Prisma v5.12.0 Release Notes](https://github.com/prisma/prisma/releases/tag/5.12.0)

## Checklist

- [x] Tests updated
- [ ] Docs updated


kentcdodds (commented):
Thanks. But I don't think this isn't any more efficient than what we have currently. Prisma's just going to turn that `createMany` into a bunch of insert statements anyway. Am I mistaken?

rajeshg:
I couldn't pinpoint the exact commit where they added it to the Prisma repository. Before 5.12.0 bulk inserts can be done only using 
`$queryRawUnsafe` method. Their release notes, definitely makes me think this should perform better than sequential writes. 

> createMany() is a method on Prisma Client, released back in version [2.16.0](https://github.com/prisma/prisma/releases/tag/2.16.0), that lets you insert multiple records into your database at once. This can be really useful when seeding your database or inserting bulk data.

In this particular case, since we are hardly making 10 inserts it shouldn't result in any tangible benefits. But this should serve as an example to others when they seed their data. 

I'm definitely seeing some perf gains when running this on a seed script locally with about 4000 records (Some of them with blobs of images and json data).

NGimbal:
Here's a relevant comment that I ran across the other day:
https://github.com/prisma/prisma/issues/10710#issuecomment-1198906656

kentcdodds (commented):
Thanks for the research! If there are perf gains, maybe we should do this for the users and notes that are created as well?

kentcdodds (approved):
Thanks!

Issue #688, opened 2024-04-07
dev-server long running command does not give back shell input correctly on fish shell


fredericrous:
I'm using `fish` shell on mac, and whenever I close the dev server, the input of the shell is messed up, I cannot see what I am writing, the up command shows `^[[A` instead of going back one line in history

basically I'm seeing this

<img width="635" alt="image" src="https://github.com/epicweb-dev/epic-stack/assets/702227/3093a6dc-207b-4e82-b8fa-5738b77fdd1a">

the issue goes away when I execute a new command. it can also be a fake command. I generally do `ls`

note: I'm using hyper terminal but I can reproduce this inside VSCode. I do not seem to get the issue with zsh

I think the issue comes from https://github.com/epicweb-dev/epic-stack/blob/d663177944e97f2705ec69482d3cc3c56865b2c8/server/dev-server.js#L9
input is ignored.. and stays ignored even after a Ctrl+C, until I execute a new command

kentcdodds:
I'm not certain this is an issue with the epic stack. Feels like an issue for execa or something wrong with your system. So I'm going to close this issue. Sorry!

fredericrous:
I'm sure all it takes to reproduce the issue is to install fish and try it for yourself. On my end no worries, I fixed the issue. I just wanted to share the solution

kentcdodds:
Thanks!

PR #689, opened 2024-04-07
dev-server writes to stdio without inherit

fredericrous:
fix https://github.com/epicweb-dev/epic-stack/issues/688

yeah I'm not a cool kid, I don't know if I can fix this issue with execa. it's pure nodejs and just works

### DIFF
--- a/server/dev-server.js
+++ b/server/dev-server.js
@@ -1,19 +1,20 @@
-import { execa } from 'execa'
+import { spawn } from 'node:child_process'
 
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks, but I'm not sure I want this solved in this way. I want to stick with execa for most of our scripts like this so I'd rather find a way to fix this with execa.

fredericrous:
yeah I'm old school. It looks like execa is not achieving its promise and I dont find it more attractive than the native api. I fail to understand its value

PR #693, opened 2024-04-10
Update examples.md to include Remix PWA

ShafSpecs:
Updated the example doc to include Remix PWA latest guide on offline apps.


kentcdodds (approved):
Thanks!

PR #696, opened 2024-04-12
sort imports, no relative imports

zwhitchcox:
* Prettier can automatically sort imports, instead of just getting a lint warning
* Don't allow relative imports - this makes refactoring more difficult as well as finding where a file is used


kentcdodds (changes requested):
I like relative imports

zwhitchcox:
Oh ok, sorry

kentcdodds (commented):
ESLint should be handling sorting of imports for us. I think I'm happy leaving it that way. Thanks anyway!

zwhitchcox:
You want to make it harder on your users?

kentcdodds:
🤔 I'm not sure what you mean. How is this harder?

zwhitchcox:
Well, they'll have to fix it manually rather than have prettier fix it automatically right? Or am I missing something? `eslint . --fix` does sort the imports, at least when I tried it.

kentcdodds:
Oh, I have eslint autofix on save. If you don't have that then yeah it'll require more work. But I recommend autofix on save.

PR #697, opened 2024-04-12
import css with vite

zwhitchcox:
With vite, we can import stylesheets directly


kentcdodds (changes requested):
I actually like doing it the way we're doing it currently. It's intentional: https://www.epicweb.dev/tips/do-not-bundle-all-your-css

Thanks anyway!

zwhitchcox:
Ahhh, I see! Yes, that is better…thanks for sharing your reasoning!

Issue #701, opened 2024-04-17
/swapon command failing when deploying to production


lewisblackburn:
The way swap files are created has recently changed on fly.io. I don't think the setup-swap.js file is needed anymore and we can just add `swap_size_mb` in `fly.toml`.

Here is the blog post by fly: https://community.fly.io/t/a-new-way-to-rootfs/19196.

kentcdodds:
Nice find! I welcome a pull request!

kinggoesgaming:
I can get it in a few hours (currently at work)

PR #706, opened 2024-04-21
build(fly): use `fly.toml` `.swap_size_mb` to create swap

kinggoesgaming:
The previous manual swap creation no longer works.

Fly changed how their VMs work with swaps. This breaks Epic Stack as it currently is. This particularly targets the `swapon` command. However they provide a `.swap_size_mb` option as the replacement in `fly.toml` now, which I have changed to in this PR.

## Test Plan

Simply deploying to the `dev` branch (when the staging deploy occurs should not fail and website should show the latest deployment when visited.

## Checklist

- [ ] Tests updated (no tests should be affected. But if I am wrong let me know)
- [x] Docs updated

## Screenshots

No visual change

closes: #701


kentcdodds (approved):
Awesome. Thank you!

PR #726, opened 2024-05-09
fix: ignore inlining small assets for the svg sprite

sorodrigo:
Vite by default will inline small assets to avoid extra requests, this wont work when using sprites. In this case we need to make sure the file is referenced as an URL otherwise the browser will block it.

See:
- https://stackoverflow.com/questions/69076708/getting-unsafe-attempt-to-load-url-dataimage-svgxml-in-safari-using-a-thr

- https://stackoverflow.com/questions/29493965/unsafe-attempt-to-load-url-svg

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -14,6 +14,12 @@ export default defineConfig({
 			external: [/node:.*/, 'stream', 'crypto', 'fsevents'],
 		},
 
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #729, opened 2024-05-10
fix: ignore generated icons folder

isaacplmann:
When an app is created with create-epic-app, the gitignore file does not include /app/components/ui/icons

## Test Plan

1. `npx create-epic-app`
2. Look at the `.gitignore` file

### DIFF
--- a/remix.init/gitignore
+++ b/remix.init/gitignore
@@ -20,3 +20,6 @@ node_modules
 
 # Easy way to create temporary files/folders that won't accidentally be added to git
 *.local.*
+
+# generated files
+/app/components/ui/icons


kentcdodds (changes requested):
This is already included: https://github.com/epicweb-dev/epic-stack/blob/69b42889b894ff952f9267931801adc1dfb46cdc/.gitignore#L25

🤔

isaacplmann:
Maybe this file is not the correct place to make the change, but when you run `npx create-epic-app` the `.gitignore` file does not have the `/app/components/ui/icons` line.

jacobparis:
Maybe we should just remove it from the .remix-init entirely if it's overwriting the real one

kentcdodds (approved):
Oops, that's right. Thanks!

PR #730, opened 2024-05-10
feat: restrict path aliases to app and test folders

isaacplmann:
If another project is ever added to the same repository, you'll want to restrict the way projects import from each other.  Creating path aliases for `/app` and `/test` allows all the current code to work in the same way it currently does, while leaving open the possibility of adding an `app2` alias in the future that is handled differently.


kentcdodds (approved):
I'm good with that. Thank you!

PR #735, opened 2024-05-17
Error Logging

zwhitchcox:
Currently, errors are not logged to the console, only to Sentry (if it's enabled)

## Test Plan

This was manually tested

## Checklist

- ~[ ] Tests updated~ -- additional tests unnecessary, as it's just logging to the console
- ~[ ] Docs updated~ -- updating docs unnecessary, as this will be the expected behavior

## Screenshots
Previously, only an 500  status was reported, but not the error itself:
Screenshot 2024-05-17 at 6 16 42 PM

We now get the full error and stack trace along with the HTTP status:
Screenshot 2024-05-17 at 7 14 17 PM

### DIFF
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -13,6 +13,7 @@ import { getEnv, init } from './utils/env.server.ts'
 import { getInstanceInfo } from './utils/litefs.server.ts'
 import { NonceProvider } from './utils/nonce-provider.ts'
 import { makeTimings } from './utils/timing.server.ts'
+import chalk from 'chalk'
 
 const ABORT_DELAY = 5000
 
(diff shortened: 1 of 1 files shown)


app/entry.server.tsx
```diff
@@ -97,9 +98,16 @@ export function handleError(
 	error: unknown,
 	{ request }: LoaderFunctionArgs | ActionFunctionArgs,
 ): void {
+	// Skip capturing if the request is aborted as Remix docs suggest
```
zwhitchcox, line 101: This is taken directly from Sentry, so the Sentry behavior won't change:

https://github.com/getsentry/sentry-javascript/blob/531779300c186f89afff0c5bad9f802b2140a325/packages/remix/src/utils/instrumentServer.ts#L151-L156

kentcdodds (approved):
Awesome. Thanks!

PR #736, opened 2024-05-18
Pr/error logging

zwhitchcox:
Fixes bug in #735, where errors would get logged 2X when the error originated in the client code.

I've removed the duplicate error logging, so now, all errors are logged, and they are only logged one time.

## Test Plan

This was manually tested.

## Checklist

- ~[ ] Tests updated~ -- additional tests unnecessary, as it's just logging to the console
- ~[ ] Docs updated~ -- updating docs unnecessary, as this will be the expected behavior

## Screenshots

Before, error thrown within React code is logged 2X:
Screenshot 2024-05-18 at 1 44 48 PM

After, they're logged only once:
Screenshot 2024-05-18 at 1 45 21 PM

### DIFF
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -7,13 +7,13 @@ import {
 } from '@remix-run/node'
 import { RemixServer } from '@remix-run/react'
 import * as Sentry from '@sentry/remix'
+import chalk from 'chalk'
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #737, opened 2024-05-20
chore: move theme switch to full stack component

lsbyerley:
If merged, this PR moves the theme switch component out of root and into it's own full stack component.

I have made this change in my own codebase after coming across https://www.epicweb.dev/full-stack-components and in an effort to reduce bloat in the root file. If this change is not wanted, feel free to close :)

## Test Plan

Functionality and display of the theme switch component should not change.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/routes/resources+/theme-switch.tsx
+++ b/app/routes/resources+/theme-switch.tsx
@@ -1,0 +1,114 @@
+import { useForm, getFormProps } from '@conform-to/react'
+import { parseWithZod } from '@conform-to/zod'
+import { invariantResponse } from '@epic-web/invariant'
+import {
+	json,
+	type ActionFunctionArgs,
+} from '@remix-run/node'
+import {
+	useFetcher,
+	useFetchers,
+} from '@remix-run/react'
+import { z } from 'zod'
+import { Icon } from '#app/components/ui/icon.tsx'
+import { useHints } from '#app/utils/client-hints.tsx'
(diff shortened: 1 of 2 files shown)


app/routes/resources+/theme-switch.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { Icon } from '#app/components/ui/icon.tsx'
+import { useHints } from '#app/utils/client-hints.tsx'
+import { useRequestInfo } from '#app/utils/request-info.ts'
+import { type Theme, setTheme } from '#app/utils/theme.server.ts'
+
+const ThemeFormSchema = z.object({
+	theme: z.enum(['system', 'light', 'dark']),
+})
+
+export async function action({ request }: ActionFunctionArgs) {
```
kentcdodds, line 22: This was copied over, but the original action in the root was not removed.

kentcdodds (changes requested):
I'm a fan. Thanks. Just one thing.

kentcdodds:
This will not negatively impact performance :)

app/routes/resources+/theme-switch.tsx
lsbyerley: ah good catch, thanks.. removed!

kentcdodds (approved):
Thanks!

PR #739, opened 2024-05-23
add RESEND_API_KEY to Dockerfile and deploy.yml

wginsberg:
This adds the `RESEND_API_KEY` env var to the `Dockerfile` and `deploy.yml`

Submitting this PR as it worked for me, I'm not sure if there was already another way to do it.

## Test Plan

Set the value of `RESEND_API_KEY` as a secret in github and deploy with Github Actions as usual.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds:
For secret environment variables that are only needed during the application runtime, you use `fly secrets` to set them. 

I hope that helps!

PR #742, opened 2024-05-25
feat: add note about multi-branch dev and schema conflicts

DennisKraaijeveld:
Decided to add some content about migrations while working with different feat/dev branches. I ran into some issues here so decided to make a small change into my workflow. You might even automate this while creating new branches, but don't think that is feasible.

## Checklist

- [x] Tests updated
- [x] Docs updated

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -144,6 +144,18 @@ to change that to `firstName` and `lastName` instead. Here's how you'd do that
 By following this strategy, you can ensure zero downtime deploys and schema
 migrations.
 
+### Multi-Branch Development
+
+When working with Prisma and SQLite in a multi-branch development environment, it's essential to handle migrations carefully to avoid conflicts and ensure a smooth development process. If migrations are applied on a feature branch but not reflected in the Prisma schema on the branch being switched to, it can lead to issues and unexpected behavior.
+
+I recommend creating a dedicated SQLite database file for each branch your work on.
+
+1. Duplicate your baseline SQLite database file. In most cases this would be the one in your `dev` branch.
+2. Rename the duplicate file with an identifier that reflects the branch name (e.g., `workflow_engine_branch_data.db`).
+3. Change the `DATABASE_PATH` value inside the `.env` file to the current branch database (e.g., `DATABASE_PATH="./prisma/workflow_engine_branch_data.db"`)
+
+
+
 ## Seeding Production
 
 In this application we have Role-based Access Control implemented. We initialize
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ node_modules
 .env
 .cache
 
-/prisma/data.db
+/prisma/*.db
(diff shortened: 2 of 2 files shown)


docs/database.md
```diff
@@ -144,6 +144,18 @@ to change that to `firstName` and `lastName` instead. Here's how you'd do that
 By following this strategy, you can ensure zero downtime deploys and schema
 migrations.
 
+### Multi-Branch Development
+
+When working with Prisma and SQLite in a multi-branch development environment, it's essential to handle migrations carefully to avoid conflicts and ensure a smooth development process. If migrations are applied on a feature branch but not reflected in the Prisma schema on the branch being switched to, it can lead to issues and unexpected behavior.
+
+I recommend creating a dedicated SQLite database file for each branch your work on.
```
kentcdodds, line 151:
```suggestion
We recommend creating a dedicated SQLite database file for each branch your work on.
```

docs/database.md
```diff
@@ -144,6 +144,18 @@ to change that to `firstName` and `lastName` instead. Here's how you'd do that
...
+
+1. Duplicate your baseline SQLite database file. In most cases this would be the one in your `dev` branch.
+2. Rename the duplicate file with an identifier that reflects the branch name (e.g., `workflow_engine_branch_data.db`).
+3. Change the `DATABASE_PATH` value inside the `.env` file to the current branch database (e.g., `DATABASE_PATH="./prisma/workflow_engine_branch_data.db"`)
```
kentcdodds, line 155: I feel like this could be very tedious to remember to update every time you change branches 🤔 can we improve.

DennisKraaijeveld: Yes, that is very true. However, the available options could be limited. You could:

1. `npx prisma migrate reset` which deletes all the data in the `data.db` and re-runs migrations. This works but you should make sure the `seed.ts file is exactly how you wanted to shape your test data. 

2. Automate this using Husky/Git Hooks and write a shell script which duplicates and prefixes the database ONLY when creating a new branch. It then updates the .env as well.

When you would switch back to an existing branch it would just update the .env and set the DB url.


I do think the question would be: Do you want just a note inside the docs with the options you have as a user. Or do you want to offer a solution for the migration errors people can run into when having multiple branches.


Reset the db and re-run `seed.ts` is fine for me too, but I can imagine not everyone is on point updating the seed script

kentcdodds: I'm just not sure I feel comfortable with this workflow personally. The problem only shows up when your branch modifies the schema. And keeping the seed up-to-date is an important practice anyway. So the workflow would be to switch the branch then run the seed script 🤷‍♂️

DennisKraaijeveld: Fair point! Then lets keep it like it is. Maybe add a note/reminder about potential conflicts and that this is why you should always keep updating your seed

kentcdodds: That sounds good to me!

PR #747, opened 2024-05-28
specify Dockerfile and .dockerignore in fly.toml

zwhitchcox:
Moving the `Dockerfile` and `.dockerignore` is not ideal, as if the process gets interrupted, they will not be moved back to the correct location.

Utilize the [dockerfile](https://fly.io/docs/reference/configuration/#specify-a-docker-image) and [ignorefile](https://fly.io/docs/reference/configuration/#specify-a-docker-ignore-file) `fly.toml` options to specify the location of the `Dockerfile` and `.dockerignore`.

## Test Plan

I tested this in my own project to make this works as expected.

## Checklist

~- [ ] Tests updated~ see above
- [*] Docs updated

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -154,12 +154,6 @@ jobs:
           file: 'fly.toml'
           field: 'app'
 
-      # move Dockerfile to root
-      - name: 🚚 Move Dockerfile
-        run: |
-          mv ./other/Dockerfile ./Dockerfile
-          mv ./other/.dockerignore ./.dockerignore
-
       - name: 🎈 Setup Fly
         uses: superfly/flyctl-actions/setup-flyctl@1.5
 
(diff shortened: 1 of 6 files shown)


zwhitchcox:
dockerignore file isn't working correctly.

kentcdodds (commented):
I'm definitely interested in this if we can get it working!

zwhitchcox:
I think this is an upstream issue with `flyctl`. I think the files are ignored, but the build process is taken over by [grpc](https://github.com/superfly/flyctl/blob/master/internal/build/imgsrc/dockerfile_builder.go#L483), and it doesn't honor the `.dockerignore` option while the files are being transferred to the build machine.

So, the files _probably_ are ignored (I assume), but you still transfer your full directory to the builder, which is ~1GB just in `node_modules`, so it's going to significantly increase the deploy time.

I opened a [community support ticket](https://community.fly.io/t/ignorefile-option-being-ignored/20096), so maybe they will be able to shed some more light.

PR #755, opened 2024-06-03
fix: authenticator uses connectionSessionStorage for errors

### LINKED ISSUE #750: Inquiry: Is the authenticator not using the connection session?
https://github.com/epicweb-dev/epic-stack/blob/7923be98dc8a3338b70619ea828f1048afe816bf/app/routes/_auth%2B/onboarding_.%24provider.tsx#L94

Im curious if this line should be trying to get the authenticator error from the connectionSessionStorage which is what is passed to the Authenticator constructor. Would the authenticator, upon failure not put the error on the connection session? How does the remix authenticator know about the auth session. 
I could be mistaken but I was under the impression that the Epic Stack does not use remix-auth directly to manage the session. But instead only pass around a user profile when the user uses the provider.

natac13:
In `utils/auth.server.ts` the `new Authenticator` is called with the `connectionSessionStorage`. However, in the `auth+/onboarding_.$provider.tsx` route the code was trying to grab the `sessionErrorKey` from the `authSession` instean of the `connnectionSession`. This fixes that.

## Test Plan

Simulate an error causd by the remix-auth authenticator failing when the GitHub (or other) social auth provider fails. 

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

fixes #750

### DIFF
--- a/app/routes/_auth+/onboarding_.$provider.tsx
+++ b/app/routes/_auth+/onboarding_.$provider.tsx
@@ -30,6 +30,7 @@ import {
 	signupWithConnection,
 	requireAnonymous,
 } from '#app/utils/auth.server.ts'
+import { connectionSessionStorage } from '#app/utils/connections.server'
 import { ProviderNameSchema } from '#app/utils/connections.tsx'
 import { prisma } from '#app/utils/db.server.ts'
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #774, opened 2024-06-18
fix nonce error with remix dev server

zwhitchcox:
When running the dev server without a nonce, e.g. via `pnpm remix vite:dev -p 3000`, you currently get an error:

Screenshot 2024-06-18 at 12 44 22 PM

## Test Plan

Tested manually

## Checklist

- ~[ ] Tests updated~ See above
- ~[ ] Docs updated~ Not necessarily, just a bug fix

## Screenshots

See above


kentcdodds (approved):
Good catch. Thank you!

Issue #744, opened 2024-05-28
Sentry really slows down requests


aon:
I came into this issue with my own application based on epic-stack and reported it here https://github.com/remix-run/remix/issues/9396. I tried just now with epic-stack and the slowness is the same.

Steps to reproduce
- Build the app with all sentry env variables set
- Run the app
- Go to pages that have `loader` defined (e.g. the users notes)
- Experience 200/300ms of load time

If the app is then run without sentry env variables, the loading time is around 10/15ms as it should be for a local app.

Do you know what might be going on here? Is there some misconfiguration or is just the cost of using sentry? I have been battling this error for weeks and tried a lot of sentry configurations but I don't seem to have an answer.

kentcdodds:
There's definitely something wrong here. You may want to open an issue on sentry

kentcdodds:
Looks like the issue you opened on sentry is getting movement, so I'll close this here. Thanks!

PR #780, opened 2024-06-23
feat: update sentry

aon:
Update Sentry to latest version, following bug detected in https://github.com/epicweb-dev/epic-stack/issues/744 and bug reported (and fixed) in Sentry:

- https://github.com/getsentry/sentry-javascript/issues/12285

## Checklist

- [x] Tests updated
- [x] Docs updated


server/index.ts
```diff
@@ -19,9 +18,9 @@ const IS_PROD = MODE === 'production'
 const IS_DEV = MODE === 'development'
 const ALLOW_INDEXING = process.env.ALLOW_INDEXING !== 'false'
 
-const createRequestHandler = IS_PROD
-	? Sentry.wrapExpressCreateRequestHandler(_createRequestHandler)
-	: _createRequestHandler
+if (IS_PROD && process.env.SENTRY_DSN) {
+	import('../app/utils/monitoring.server.ts').then(({ init }) => init())
```
kentcdodds, line 22: This won't work when deployed because this file isn't bundled. Can we move the monitoring file to the server directory?

kentcdodds (changes requested):
Thank you!

server/index.ts
aon: Fixed!

kentcdodds (approved):
Solid. Thank you!

Issue #682, opened 2024-03-30
Rate limiting can be bypassed with a spoofed ip


nickjamesio:
First off, thanks for creating the epic stack! I'm learning about rate limiting so I thought I'd use epic-stack for reference. While looking over the repo, these two lines caught my attention: 

- https://github.com/epicweb-dev/epic-stack/blob/800022dae82530dd48ef0ba35abe666eac1c619e/server/index.ts#L152
- https://github.com/epicweb-dev/epic-stack/blob/800022dae82530dd48ef0ba35abe666eac1c619e/server/index.ts#L39

I was curious to see if it is indeed impossible to spoof `X-Forwarded-For`. I came across this forum post https://community.fly.io/t/recommended-setting-for-trust-proxy-on-express/6346 and I tried the curl request they provided and it looks like you can indeed spoof `X-Forwarded-For`.

Request for spoofing IP
`curl -sv\
    -H "Host: injected.host"\
    -H "X-Forwarded-Proto: injected"\
    -H "X-Forwarded-For: 1.2.3.4"\
    -H "X-Forwarded-Port: 1234"\
    -H "X-Forwarded-Ssl: maybe"\
    https://debug.fly.dev/`

While looking at express-rate-limit, it looks like it uses `req.ip` as the unique identifier. See [here](https://github.com/express-rate-limit/express-rate-limit/blob/fe37463bb24269e1963c340da72f123158f4d4b6/source/lib.ts#L220)

The express docs have this to say about req.ip

> The [req.ip](https://expressjs.com/en/api.html#req.ip) and [req.ips](https://expressjs.com/en/api.html#req.ips) values are populated based on the socket address and X-Forwarded-For header, starting at the first untrusted address.

I ran the epic stack locally and I could indeed get past the rate limiting with a spoofed IP. See [this loom video](https://www.loom.com/share/8c61315930d74f4dbaa37423cee9f284)

Based on this info, I am pretty sure rate limiting can be bypassed when an epic stack application is hosted in fly.io I think one solution would be to use the `Fly-Client-Ip` header since it would appear that header cannot be spoofed.

kentcdodds:
Nice find! I welcome a PR to switch our implementation to use the `Fly-Client-Ip` header!

nickjamesio:
I will def work on it!

andrecasal:
Great work Nick!

nickjamesio:
[...]

PR #782, opened 2024-06-25
Use Fly-Client-Ip as rate limit key

nickjamesio:
The ip used for rate limiting can be spoofed as detailed here https://github.com/epicweb-dev/epic-stack/issues/682

### Test Plan
In the browser, submit to a protected route such as login to trigger the rate limit.
Use curl or Postman to submit to the login endpoint but attempt to override the Fly-Client-Ip header for each request. Rate limiting should still be triggered

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -146,9 +146,14 @@ const rateLimitDefault = {
 	max: 1000 * maxMultiple,
 	standardHeaders: true,
 	legacyHeaders: false,
-	// Fly.io prevents spoofing of X-Forwarded-For
-	// so no need to validate the trustProxy config
 	validate: { trustProxy: false },
+	// Malicious users can spoof their IP address which means we should not deault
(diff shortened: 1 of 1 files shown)


nickjamesio:
I don't actually use the Epic stack and I had trouble deploying this to Fly to verify the fix :( This is the error I kept getting. 

```
Error: failed to fetch an image or build from source: error building: failed to receive status: rpc error: code = Unavailable desc = error reading from server: remote error: tls: bad record MAC
```
It still runs locally though. I copied some of the code I used so it should work. I'm hoping you can try it out in Fly

kentcdodds (approved):
Perfect. Thank you!

kentcdodds:
Could you fix the TypeScript issue?

nickjamesio:
> Could you fix the TypeScript issue?

Sure can! Fixing now

kentcdodds (approved):
Awesome, thank you!

PR #784, opened 2024-06-29
remove redundant sourcemap flags

nichtsam:
## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "build": "run-s build:*",
     "build:icons": "tsx ./other/build-icons.ts",
-    "build:remix": "remix vite:build --sourcemapClient --sourcemapServer",
+    "build:remix": "remix vite:build",
     "build:server": "tsx ./other/build-server.ts",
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks! Have you verified everything still works as expected without these options?

kiliman:
I'm pretty sure you need those flags to get sourcemaps on a production build. I'm not sure what is meant by "redundant".

nichtsam:
I'm not sure how to verify this, here are what I checked:

I built 
- with flags and without `build.sourcemap = true` => generates sourcemap
- without flags and with `build.sourcemap = true` => generates sourcemap
- withboth => generates sourcemap
- without both => no sourcemap

I went to check the source code of `remix vite:build`, and I think what the flags do is setting `build.sourcemap = true` for their own environment, which is the same as setting it in `vite.config.ts`.
https://github.com/remix-run/remix/blob/main/packages/remix-dev/vite/build.ts#L229

kentcdodds:
Thanks!

PR #787, opened 2024-07-02
Fix hydration warning on script nonces

aon:
This PR just fixes the annoying hydration warning that appears on Firefox on script nonces:

image

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -181,6 +181,7 @@ function Document({
 					dangerouslySetInnerHTML={{
 						__html: `window.ENV = ${JSON.stringify(env)}`,
 					}}
+					suppressHydrationWarning
 				/>
 				<ScrollRestoration nonce={nonce} />
 				<Scripts nonce={nonce} />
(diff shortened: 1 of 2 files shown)


kentcdodds:
I can't reproduce this issue with https://www.epicstack.dev using firefox.

image

aon:
I can't either, strange. But on dev mode the message is indeed there. Can you try it on a local environment?

kentcdodds:
Ah yes, I can reproduce it locally. I'll have to look into it a bit more later. I don't like using this attribute because it just hides the problem, but it may not fix the problem.

aon:
It does not indeed fix the problem. I've been reading your issues raised on the React repository and there doesn't seem to be a good solution so far. I implement this on my own projects because I couldn't think of any hydration issue that could arise on these scripts, and to prevent having errors in the console tab. Although I do agree with your statement above.

kentcdodds:
More on this here: https://github.com/epicweb-dev/epic-stack/discussions/768

Issue #791, opened 2024-07-11
Issue with `db-utils/cleanupDB` function when actual foreign key constraints exists


ovistoica:
I added some tables to the database and after inserting data in, the DB couldn't clean anymore given foreign key constraints. 

This was weird since cleanupDb set `PRAGMA foreight_keys = off` in the transaction: 

```ts
export async function cleanupDb(prisma: PrismaClient) {
  const tables = await prisma.$queryRaw<
    { name: string }[]
  >`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' AND name NOT LIKE '_prisma_migrations';`

  await prisma.$transaction([
    // Disable FK constraints to avoid relation conflicts during deletion
    prisma.$executeRawUnsafe(`PRAGMA foreign_keys = OFF`),
    // Delete all rows from each table, preserving table structures
    ...tables.map(({ name }) =>
      prisma.$executeRawUnsafe(`DELETE from "${name}"`),
    ),
    prisma.$executeRawUnsafe(`PRAGMA foreign_keys = ON`),
  ])
}
```

The issue, according to chatGPT, is this: 

> The error occurs because the PRAGMA foreign_keys = OFF statement within the prisma.$transaction block does not persist across the individual prisma.$executeRawUnsafe statements. Each prisma.$executeRawUnsafe execution is a separate statement, and SQLite's PRAGMA commands do not affect the behavior of other statements outside of the immediate execution context.

> Here’s an updated approach to ensure that the PRAGMA foreign_keys = OFF setting persists while deleting the rows:

> Execute the PRAGMA foreign_keys = OFF and PRAGMA foreign_keys = ON statements outside the transaction to ensure they are applied properly.
> Use a single transaction to delete the data from all tables.
> Here's a revised version of your code:
```ts
const tables = await prisma.$queryRaw<
  { name: string }[]
>`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' AND name NOT LIKE '_prisma_migrations';`

// Disable FK constraints to avoid relation conflicts during deletion
await prisma.$executeRawUnsafe(`PRAGMA foreign_keys = OFF`)

// Delete all rows from each table, preserving table structures
await prisma.$transaction([
  ...tables.map(({ name }) =>
    prisma.$executeRawUnsafe(`DELETE FROM "${name}"`),
  )
])

// Re-enable FK constraints
[...]

PR #795, opened 2024-07-12
Fix `cleanupDb` when foreign key constraints exist

ovistoica:
PR to address https://github.com/epicweb-dev/epic-stack/issues/791 
## Test Plan

If the `npm run setup` works as expected, the seeding and the db cleanup works correctly. 

To test this fix fully, we'd need to add some data to the database with FK constraints, not have `ON DELETE CASCADE` on it, and expect `cleanupDb` to succeed.

Also, another good test would be to monkey patch the `cleanupDb` to throw inside the transaction and expect `foreign_keys` to be still on (catch clause reached)

## Checklist

- [x] Tests updated - no tests need updating
- [x] Docs updated - no need

### DIFF
--- a/tests/db-utils.ts
+++ b/tests/db-utils.ts
@@ -121,9 +121,9 @@ export async function cleanupDb(prisma: PrismaClient) {
 	>`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' AND name NOT LIKE '_prisma_migrations';`
 
 	try {
+		// Disable FK constraints to avoid relation conflicts during deletion
 		await prisma.$executeRawUnsafe(`PRAGMA foreign_keys = OFF`)
 		await prisma.$transaction([
-			// Disable FK constraints to avoid relation conflicts during deletion
 			// Delete all rows from each table, preserving table structures
 			...tables.map(({ name }) =>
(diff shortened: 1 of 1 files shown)


tests/db-utils.ts
```diff
@@ -120,13 +120,19 @@ export async function cleanupDb(prisma: PrismaClient) {
...
+			...tables.map(({ name }) =>
+				prisma.$executeRawUnsafe(`DELETE from "${name}"`),
+			),
+		])
+		await prisma.$executeRawUnsafe(`PRAGMA foreign_keys = ON`)
+	} catch (error) {
+		console.error('Error cleaning up database:', error)
+		// Ensure FK constraints are re-enabled
+		await prisma.$executeRawUnsafe(`PRAGMA foreign_keys = ON`)
+	}
```
kentcdodds, line 137: Can simplify a bit

```suggestion
	} catch (error) {
		console.error('Error cleaning up database:', error)
	} finally {
		await prisma.$executeRawUnsafe(`PRAGMA foreign_keys = ON`)
	}
```

kentcdodds (approved):
Looks great. Thank you for taking the time to contribute back!

tests/db-utils.ts
ovistoica: Good point!

PR #796, opened 2024-07-12
fix expires date in playwright addCookies

onemen:
There are type difference between set-cookie-parser and playwright cookies.

`expires` in cookie-parser is `Date | undefined`
playwright addCookies use `number | undefined'`

This PR convert the Date to number using `getTime`


kentcdodds (approved):
Thank you!

PR #797, opened 2024-07-15
allow connection to hmr server locally

sys13:
Vite's hot module reload server is on this port. This opens the CSP to connect to it in dev.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -115,7 +115,7 @@ app.use(
 			reportOnly: true,
 			directives: {
 				'connect-src': [
-					MODE === 'development' ? 'ws:' : null,
+					...(MODE === 'development' ? ['ws:', 'localhost:24678'] : []),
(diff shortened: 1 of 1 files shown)


server/index.ts
```diff
@@ -115,7 +115,7 @@ app.use(
 			reportOnly: true,
 			directives: {
 				'connect-src': [
-					MODE === 'development' ? 'ws:' : null,
+					...(MODE === 'development' ? ['ws:', 'localhost:24678'] : []),
```
kentcdodds, line 118: Is there any other way for us to determine this websocket address? The problem is, what happens if I try to run my app two times at once?

kentcdodds (commented):
Thanks for the contribution! The CSP is set to report only, so it shouldn't affect functionality and I've not personally seen any CSP warnings either. Can you screenshot your console with warnings?

Also, I'm concerned about hard coding the port.

Issue #792, opened 2024-07-11
Use `styleText` from `node:util`


kinggoesgaming:
It provides the same functionality and one less dependency to worry about.

kentcdodds:
I'm definitely up for a PR for this.

kinggoesgaming:
If no one gets to this by Saturday, I'll do it

PR #798, opened 2024-07-21
build(deps): remove `chalk` and use `styleText` instead

kinggoesgaming:
- **refactor: replace `chalk` with `styleText`**
- **build(deps): remove `chalk`**
- **style(imports): fix import order**

## Test Plan

## Screenshots

No change that requires diff of screenshots

closes: #792

### DIFF
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -1,5 +1,5 @@
-import { PassThrough } from 'stream'
 import { styleText } from 'node:util'
+import { PassThrough } from 'stream'
 import {
 	createReadableStreamFromReadable,
 	type LoaderFunctionArgs,
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -1,4 +1,5 @@
 import { PassThrough } from 'stream'
+import { styleText } from 'node:util'
 import {
 	createReadableStreamFromReadable,
 	type LoaderFunctionArgs,
@@ -7,7 +8,6 @@ import {
 } from '@remix-run/node'
 import { RemixServer } from '@remix-run/react'
 import * as Sentry from '@sentry/remix'
-import chalk from 'chalk'
 import { isbot } from 'isbot'
 import { renderToPipeableStream } from 'react-dom/server'
 import { getEnv, init } from './utils/env.server.ts'
@@ -102,7 +102,7 @@ export function handleError(
 		return
 	}
 	if (error instanceof Error) {
-		console.error(chalk.red(error.stack))
+		console.error(styleText('red', String(error.stack)))
 		void Sentry.captureRemixServerException(
 			error,
 			'remix.server',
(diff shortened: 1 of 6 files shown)


app/entry.server.tsx
```diff
@@ -1,4 +1,5 @@
 import { PassThrough } from 'stream'
+import { styleText } from 'node:util'
```
kinggoesgaming, line 2: I am planning on a future PR that makes sure all builtin imports are prefixed with `node:`, so the ESLint warning for import order should be fine (unless you want it fixed

index.js
```diff
@@ -1,6 +1,8 @@
 import 'dotenv/config'
+// 👆 this needs to imported first
+
+import { styleText } from 'node:util'
```
kinggoesgaming, line 4: Same as above

app/entry.server.tsx
kentcdodds: Let's fix it

kentcdodds (changes requested):
Super!

jacobparis:
styleText doesn't land until Node 22 (node 21.7 technically) which won't be LTS until October, so I don't think we can merge this until then based on this decision doc

https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/021-node-version.md

kinggoesgaming:
> styleText doesn't land until Node 22 (node 21.7 technically) which won't be LTS until October, so I don't think we can merge this until then based on this decision doc
> 
> https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/021-node-version.md

I am so used to using it that I didn't realize it is still in `Active development` stage.

I will try to keep this PR up-to-date until it is stablized.

kinggoesgaming:
Another point: this does not actually remove `chalk` from the dependency tree, as other dependencies do depend on `chalk`, but now it is a transitive `dev` dependency.

alcpereira:
> styleText doesn't land until Node 22 (node 21.7 technically) which won't be LTS until October, so I don't think we can merge this until then based on this decision doc
> 
> https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/021-node-version.md

What do you think about updating the requirement to [Node `20.12.0`](https://nodejs.org/en/blog/release/v20.12.0#text-styling)? 

I'm used to work with `.nvmrc` or `.node-version` files so that Node version manager like [fnm](https://github.com/Schniz/fnm) can select the right version for the project. 
Is it worth opening a PR to add one of those (with corresponding documentation)? So it might also unblock this PR.

kentcdodds:
Why don't node managers reference the engines config in the package.json? I'd really rather avoid adding an extra file to the root just for the node version.

kentcdodds:
I'm going to close this for now. We can revisit when Node 22 is our base.

kentcdodds:
Now we're running on node version 22. So we can look into doing this. There have been a number of changes to these different files. So if you would like to update your pull request with what is latest on main and get everything ready to go then that would be really appreciated. And then we can get this merged in. Thanks.

kinggoesgaming:
I can look into this in a few days...

kinggoesgaming:
superceded by #928

Issue #799, opened 2024-07-21
Use `.nvmrc` file to declare the `node` version to use


kinggoesgaming:
This will make it so we don't need to change the version 6 times (1 time in `package.json`, 1 in `other/Dockerfile` and 4 times in `.github/workflows/deploy.yml`. Instead it will be only 3 times.

Also anyone using `nvm` to manage `node` version can use `nvm use` to match version used with the stack for eliminating issues that can raise due to differing `node` versions (These should be rare, but still a possibility.)

kinggoesgaming:
closing this... as using `package.json` was decided in #801

PR #801, opened 2024-07-21
build(node): use `.nvmrc` to declare `node` version

kinggoesgaming:
## Test Plan

Current tests should pass as-is

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

This does not have any changes that need screenshots

closes: #799


kentcdodds (commented):
Thanks. I don't use nvm though. Is there no way to have these tools reference the engines config already in the package.json?

kinggoesgaming:
There might be a GitHub Action for fetching the value (similar to how we fetch fly app name). However I am not familiar with that. As per `.nvmrc`, node GitHub Action supports reading the version from file, I just used `.nvmrc`, as `nvm` is the most popular version manager for node.

kentcdodds:
Is there any way for these tools to reference the engines config in the package.json so we don't need an entirely separate file for managing the node version?

kentcdodds:
That all sounds fine to me. I'm fine with documentation that recommends `fnm` as long as we don't need to add yet another root-level file to the codebase.

kinggoesgaming:
> Regarding the action [setup-node](https://github.com/actions/setup-node), it also accepts `package.json` as `node-version-file` 🎉

Interesting I didn't know this.. I will update to use it

kentcdodds:
Closing this for now. Feel free to open a new PR for supporting the package.json

Issue #802, opened 2024-07-21
Remove `installGlobals` calls


kinggoesgaming:
The globals added by `installGlobals` are now natively available with `node` 20+

Ref: https://github.com/remix-run/remix/issues/7067#issuecomment-2225843886

kinggoesgaming:
Ok I think I might have misunderstood that comment. Need a second opinion before I anything here

kinggoesgaming:
> Ok I think I might have misunderstood that comment. Need a second opinion before I anything here

Looking further.. I am correct on the fact that `installGlobals` is no longer needed

kentcdodds:
Yeah, we can remove that I think

PR #803, opened 2024-07-24
build(remix): don't use `installGlobals`

kinggoesgaming:
The globals that were installed with `installGlobals` are now provided by `node` itself.

The one major change is removing `getSetCookie` helper method that was used as a hack.

## Test Plan

The current tests should pass.

## Checklist

- [x] Tests updated

## Screenshots

No change that requires screenshot change.

closes: #802


kentcdodds (approved):
Brilliant! Hooray for built-in standards!

PR #808, opened 2024-07-30
Improve docker ignorefile honoring

agcty:
`.dockerignore` files are only honored at project root, this PR changes the name to `Dockerfile.dockerignore` which makes docker actually honor the given patterns no matter on which path the ignorefile lives.

Src: https://stackoverflow.com/questions/71359019/where-to-place-dockerignore-for-docker-build-f-path-to-docker-dockerfile

Right now if you run the following command:

`docker build -t epic-stack:latest -f other/Dockerfile .`

the patterns in `.dockerignore` are in fact not ignored.

In addition, fly allows you to configure the docker and ignorefile in the build section of the configuration which means you don't need to move the files to the root anymore for the honoring of the ignore patterns happening.

Overall, this PR removes unneeded code and workarounds that are better solved with the correct configuration.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -156,12 +156,6 @@ jobs:
           file: 'fly.toml'
           field: 'app'
 
-      # move Dockerfile to root
-      - name: 🚚 Move Dockerfile
-        run: |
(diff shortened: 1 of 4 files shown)


kentcdodds (approved):
I'm much happier with this. Thank you!

PR #810, opened 2024-08-03
add util for handling remix headers generally

nichtsam:
It's a bit tedious to manually merge Remix headers since the process is mostly the same.
This PR adds a utility to streamline that process.

Related Discussion #809 
More details on the utility: https://gist.github.com/nichtsam/bff3a0e58f7edfad3985e9322582f67f

## Test Plan

- There are test cases for various functionalities.
- Test cases with real use cases ensure the utility remains practical.

## Checklist

- [x] Tests updated
- [x] Docs updated


nichtsam:
Just for the record, I changed the PR into draft because there's an issue with importing that works in production but not in development.

nichtsam:
> Just for the record, reason for switching to draft: [privatenumber/tsx#627](https://github.com/privatenumber/tsx/issues/627) This is affecting `vite` and causing imports behavior to differentiate between `tsx` and `node`.

I switch to another ESM cache-control parser to avoid encountering this issue.

kentcdodds:
I think React Router v7 (single fetch specifically) changes how headers work a fair bit so I don't think I'll be merging this since I want to get on RR7 soon. Sorry I didn't get to you on this sooner!

nichtsam:
> I think React Router v7 (single fetch specifically) changes how headers work a fair bit so I don't think I'll be merging this since I want to get on RR7 soon. Sorry I didn't get to you on this sooner!

That’s totally fine! I was also thinking that single fetch might impact things. 
Just wanted to finish up the PR. No worries, and thanks for the update!

kentcdodds:
RR7 has been merged! Feel free to check whether this is necessary and update things if so. Thanks!

app/utils/remix.server.ts
```diff
@@ -0,0 +1,103 @@
+import { type CacheControlValue, parse, format } from '@tusbar/cache-control'
```
kentcdodds, line 1: I think calling this file `remix.server.ts` may not be the right call anymore. Maybe this should just be called `headers.server.ts`.

Also, could you add some jsdoc comments to these utilities? Thanks!

app/utils/remix.server.test.ts
```diff
@@ -0,0 +1,41 @@
+import { format, parse } from '@tusbar/cache-control'
+import { describe, expect, test } from 'vitest'
+import { getConservativeCacheControl } from './remix.server.ts'
+
+describe('getConservativeCacheControl', () => {
```
kentcdodds, line 5: I prefer to not use `describe` because it's unnecessary. Just put the test blocks at the root of the file.

kentcdodds (changes requested):
Thanks! Just two changes.

PR #812, opened 2024-08-04
chore: remove unused dep @remix-run/serve

alcpereira:
While looking at logging solutions for Remix, I naturally checked how Epic Stack was doing it.
I noticed the custom implementation for serving the Remix app in production, and noticed that `@remix-run/serve` was not used, so I removed it.

Let me know if there is a reason that it's still there that I might have missed :) 

## Checklist

- N/A Tests updated
- N/A Docs updated


kentcdodds (approved):
Yep, this is no longer in use. Thanks!

PR #820, opened 2024-08-12
fix: post converted to get by proxy

### LINKED ISSUE #818: POST requests turn into GET
Hey all, I'm working on a chatbot (with Botpress Cloud) that accesses my information via API routes. The problem is that the service I'm using, goes through a proxy. When it gets to my express server, the server receives the request as an HTTP request instead of an HTTPS and the middleware set by default on the express server performs the redirect to HTTPS but loses the request method.

So TLDR; 

POST request -> Express Middleware -> GET request

```javascript
app.use((req, res, next) => {
	const proto = req.get('X-Forwarded-Proto')
	const host = getHost(req)
	if (proto === 'http') {
		res.set('X-Forwarded-Proto', 'https')
		res.redirect(`https://${host}${req.originalUrl}`)
		return
	}
	next()
})
```

I've tried doing a couple of things I've found around by adding codes `308` and `307` but then the request gets lost and the chatbot ends with network errors (error shown: `AxiosError: Network Error`). 

```javascript
if (req.method === 'POST') {
	res.redirect(307, `https://${host}${req.originalUrl}`)
	return
}
```

I'm not sure if this is the right place to ask for suggestions so I'd appreciate guidance on how to make things work with this.

CodingAleCR:
Fixes #818 , where POST requests get converted to GET requests on certain incoming requests coming from proxy (eg. Botpress)

## Test Plan

- Tested with Botpress to check if the request is received properly.
- Tested the rest of my project to review that everything was working fine.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -38,6 +38,8 @@ app.set('trust proxy', true)
 
 // ensure HTTPS only (X-Forwarded-Proto comes from Fly)
 app.use((req, res, next) => {
+	if (req.method !== 'GET') return next()
+	
 	const proto = req.get('X-Forwarded-Proto')
 	const host = getHost(req)
 	if (proto === 'http') {


kentcdodds (approved):
Thank you!

PR #811, opened 2024-08-03
refactor(imports): use `node:` protocol for builtins

kinggoesgaming:
I went through the files to ensure I didn't miss any. However with the scope of the project, I could have missed some imports. If you see any let me know.

Probably best to not immediately merge this to ensure I didn't miss any import that can be updated.
## Test Plan

closes: #800

PR #833, opened 2024-08-27
refactor(imports): missing `node:` protocol for builtins

alcpereira:
Following up on #811 , there was some missing files modifications.

I'm not sure about the change in `.vscode/settings.json`, let me know if I should undo those changes.

## Checklist

- N/A Tests updated
- N/A Docs updated


kentcdodds (approved):
Thanks!

PR #855, opened 2024-09-20
New form fields

braginteractive:
Select, Time, Number, and Combobox. 

Props might need some work. 

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (changes requested):
Thanks! I'm not certain I want this included since they're not in use. But an [example](https://github.com/epicweb-dev/epic-stack/blob/main/docs/examples.md) that has these and uses them would be welcome!

devneill:
These would be super useful as an example!

I would have really appreciated this when I was adding my own Select

PR #873, opened 2024-10-10
Catch errors that cause shutdowns and send them to Sentry

### LINKED ISSUE #872: epic-stack by default doesn't seem to report unhandled rejections to Sentry.
We've been running into some issues with Sentry reporting in our epic-stack based application, and I think the issue is present in the base template - epic-stack calls `closeWithGrace` in its entrypoint:

https://github.com/epicweb-dev/epic-stack/blob/b19ac20b756707831cab51dd6a096a655f5d76a7/index.js#L21-L27

Which will register a `process.on('unhandledRejection'` error and exit before any other unhandledRejection listeners have time to respond, including Sentry's unhandledRejectionIntegration. The result is that unhandled rejections that cause the server to crash also are not reported to Sentry, because `closeWithGrace` calls `process.exit(1)` as soon as they are received by the first registered handler.

tmcw:
- Fixes #872 

Before this PR, epic-stack has two calls to close-with-grace. This removes the redundant call, which was usually the one that ran first and prevented the other one from running. It then adds integration with Sentry so that errors that cause the server to shut down are sent to Sentry, and we give Sentry 500 milliseconds to flush any other errors that haven't been sent yet.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,14 +18,6 @@ sourceMapSupport.install({
 	},
 })
 
-closeWithGrace(async ({ err }) => {
-	if (err) {
-		console.error(chalk.red(err))
-		console.error(chalk.red(err.stack))
-		process.exit(1)
-	}
-})
-
 if (process.env.MOCKS === 'true') {
 	await import('./tests/mocks/index.ts')
 }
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Excellent. Thank you!

PR #874, opened 2024-10-11
Update deployment.md

hagenek:
Fly.io strongly recommend having 2 volumes, and I encountered issues because I only had one with no resources available.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -84,8 +84,8 @@ Prior to your first deployment, you'll need to do a few things:
    sure you change the `primary_region` in fly.toml as well):
 
    ```sh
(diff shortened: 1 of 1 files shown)


hagenek:
I actually think the fly.io free tier has been removed? Seems like this when I read here:

https://fly.io/docs/about/pricing/

lewisblackburn:
For users on the Legacy Hobby Plan, there will be no impact; however, new users will be required to use a pay-as-you-go model. To be safe, I think we should still use a single volume until the project is ready for production to keep costs as low as possible.

kentcdodds:
An additional note is that if your bill is less than $5 a month they will forgive it. So technically no free tier but practically you can host apps on fly for free if they are low use.

PR #875, opened 2024-10-12
draft: upgrade to react 19

jacobparis:
React 19 is still RC so we shouldn't merge this, but just getting a feel for what a React 19 epic stack looks like. Not much changes!
- forwardRef is used by the shadcn components, which we can remove. I assume eventually they'll start shipping those without forwardRef
- no need for meta export anymore, we can move title and description directly into JSX


sbaechler:
React 19 has been released.

kentcdodds:
The Epic Stack was upgraded to React 19 last week. https://github.com/epicweb-dev/epic-stack/releases/tag/2024-12-09

PR #876, opened 2024-10-13
Enabling Single Fetch

hakimLyon:
This PR enables the [v3_singleFetch](https://remix.run/docs/en/main/guides/single-fetch#enabling-single-fetch) feature.


kentcdodds:
Thank you! Could you debug the playwright failures please?

app/routes/_auth+/auth.$provider.ts
```diff
@@ -1,14 +1,10 @@
...
 import { authenticator } from '#app/utils/auth.server.ts'
 import { handleMockAction } from '#app/utils/connections.server.ts'
 import { ProviderNameSchema } from '#app/utils/connections.tsx'
 import { getReferrerRoute } from '#app/utils/misc.tsx'
 import { getRedirectCookieHeader } from '#app/utils/redirect-cookie.server.ts'
 
-export async function loader() {
-	return redirect('/login')
-}
-
```
kentcdodds, line 11: Why remove this?

app/root.tsx
```diff
@@ -1,5 +1,5 @@
 import {
-	json,
+	data as dataResponse,
```
kentcdodds, line 2: I think I'd prefer to just import this as `data` everywhere. I realize this means we get variable shadowing in some places, but I don't see that as being a practical concern. Those variables could (probably should) be changed.

app/routes/users+/index.tsx
```diff
@@ -41,11 +45,14 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
 	const result = UserSearchResultsSchema.safeParse(rawUsers)
 	if (!result.success) {
-		return json({ status: 'error', error: result.error.message } as const, {
-			status: 400,
-		})
+		return dataResponse(
+			{ status: 'error', error: result.error.message } as const,
+			{
+				status: 400,
+			},
```
kentcdodds, line 52:
```suggestion
			{ status: 400 },
```

index.js
```diff
@@ -1,8 +1,9 @@
 import 'dotenv/config'
 import * as fs from 'node:fs'
-import chalk from 'chalk'
-import closeWithGrace from 'close-with-grace'
+
+import { installGlobals } from '@remix-run/node'
 import sourceMapSupport from 'source-map-support'
+installGlobals({ nativeFetch: true })
```
kentcdodds, line 6: This should not be necessary with Node 20. Why is this added?

kentcdodds (changes requested):
Thanks for this! I have a few notes.

app/routes/_auth+/auth.$provider.ts
hakimLyon: Playwright failed when the login redirect was present, so I had to remove it.

index.js
hakimLyon: You're right, I just remembered that

vite.config.ts
```diff
@@ -41,6 +47,7 @@ export default defineConfig({
 			: remix({
 					ignoredRouteFiles: ['**/*'],
 					serverModuleFormat: 'esm',
+					future: { v3_singleFetch: true },
```
davidesigner, line 50: Hi, I think there's a huge problem with this.

The code is:

```ts
process.env.NODE_ENV === 'test'
  ? null
  : remix({
    future: { v3_singleFetch: true },
    // ...
  });
```

So, in the test environment, we're going to test something that isn't identical to the production or development environments.

I saw the problem because in the development environment, **MSW mocks don't work with `singleFetch` enabled**. But they do in test mode.

I think this is an issue we need to resolve before enabling `singleFetch` functionality. I don't have a solution at the moment.

kentcdodds: > MSW mocks don't work with singleFetch enabled.

Hmmm... Can you be more descriptive? What is not working? What output do you get?

@kettanaito, have you tried a Remix app with `singleFetch` enabled with MSW?

kettanaito: @kentcdodds, I don't believe I have.

I'd love to hear more info about any errors or other behaviors MSW shows when the single fetch mode is enabled. 

Afaik, single fetch should have no effect for MSW. It groups loader resource calls for nested routes, but has no effect on the requests performed within the loaders. Then, it groups all the loader resources into, well, a single fetch, and uses a "stream fetch" to send it back to the browser. That encoded, streamed response is not something you should be intercepting. Correct me if I'm wrong about that.

kettanaito: Following the [Debugging runbook](https://mswjs.io/docs/runbook) can be helpful to figure out what and why doesn't work when intercepting requests. @davidesigner, could you please let me know how far you get with that runbook?

davidesigner: Wow I realised why I had this problem, it's because I had added: `installGlobals({ nativeFetch: true });` (which wasn't necessary). When I remove this line from the `vite.config.ts` file, everything works fine.

Sorry for wasting your time...

itsmegood:
Will it be possible to have a codemod for this upgrade. Manually upgrading all this for custom/ extended project seems a bit difficult.

kentcdodds:
Maybe @jacobparis's tool could help.

vite.config.ts
kettanaito: No time wasted! I'm glad to hear that it works as expected now. Great find!

itsmegood:
> Maybe @jacobparis's tool could help.

I found this https://codemod.com/registry/remix-single-fetch-recipe by jacobparis

itsmegood:
Any update regarding this @kentcdodds

bkilrain:
I'm in the middle of the same single fetch effort on an app based off this template. I hit some funkiness with how the toasts work for error responses so came over here to see how y'all are handling it. 

The new behavior is that loaders don't rerun by default when an action returns a 4xx/5xx response. This means the root loader doesn't process the toast headers until the next GET/successful POST. 

@hakimLyon have you noticed if the toasts still work in this case?

hakimLyon:
> @hakimLyon have you noticed if the toasts still work in this case?

Hi @bkilrain , I haven’t noticed this behavior on my side. But in Remix Docs about [Submission Revalidation Behavior
](https://remix.run/docs/en/main/guides/single-fetch#submission-revalidation-behavior) It suggests using options like shouldRevalidate to customize this behavior and force loaders to rerun if needed.

bkilrain:
> > @hakimLyon have you noticed if the toasts still work in this case?
> 
> Hi @bkilrain , I haven’t noticed this behavior on my side. But in Remix Docs about [Submission Revalidation Behavior ](https://remix.run/docs/en/main/guides/single-fetch#submission-revalidation-behavior) It suggests using options like shouldRevalidate to customize this behavior and force loaders to rerun if needed.

Yep that's what I ended up doing on my side... just raising the possibility that this PR might need it as well.

```
export const shouldRevalidate: ShouldRevalidateFunction = ({
  defaultShouldRevalidate,
  actionStatus,
}) => {
  const status = String(actionStatus)
  if (status.startsWith('4') || status.startsWith('5')) {
    // 4xx & 5xx error response should be revalidated in case they contain toast headers
    return true
  }
  return defaultShouldRevalidate
}
```

kentcdodds:
Thank you so much for all your contributions to this and the #897 PR! Merged that other one so we're all set!

PR #880, opened 2024-10-30
Update Dockerfile to use npm ci

theskillwithin:
https://docs.npmjs.com/cli/v10/commands/npm-ci

> This command is similar to [npm install](https://docs.npmjs.com/cli/v10/commands/npm-install), except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

### DIFF
--- a/other/Dockerfile
+++ b/other/Dockerfile
@@ -15,7 +15,7 @@ FROM base as deps
 WORKDIR /myapp
 
 ADD package.json package-lock.json .npmrc ./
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
I feel like I've tried this before and something went wrong, but maybe it was somewhere else. Pretty sure this is fine. Thanks!

Issue #881, opened 2024-10-31
Error: Prisma [Github Actions] Migrate has detected that the environment is non-interactive. It is recommended to run this command in an interactive environment.


MoSattler:
![Screenshot 2024-10-31 at 18 17 28](https://github.com/user-attachments/assets/417965aa-7d84-42c2-9205-1762a87ea652)

It looks like the new seed command fails in GitHub actions.

PR #882, opened 2024-10-31
Update deploy.yml

MoSattler:
Closes https://github.com/epicweb-dev/epic-stack/issues/881


kentcdodds (approved):
Thanks!

I've also deleted the cache in actions to make sure this step runs so we know for sure it's working.

PR #894, opened 2024-12-17
remove glob for source maps

zwhitchcox:
sentry already uses a glob, and if you glob before generating the source maps, they will not be deleted, because they did not exist at the time of the glob

## Test Plan
We tested on our server


kentcdodds (changes requested):
Nice! Thank you for this. Can you remove the import?

zwhitchcox:
Oops, sorry, didn't catch that!

kentcdodds (approved):
Fabulous. Thanks!

PR #896, opened 2024-12-26
make use of epic-web config reset

nichtsam:
`@epic-web/config/reset.d.ts` was not used, and `CSSProperties` was not patched.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

PR #898, opened 2025-01-11
handle not found post requests

nichtsam:
not found post requests creates noice on sentry:
```You made a POST request to "/some/path" but did not provide an `action` for route "routes/$", so there is no way to handle the request.```

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
That makes sense. Thanks!

PR #903, opened 2025-01-17
change sentry import to correct package

nichtsam:
These are server-side setup and should use `@sentry/node` instead of `@sentry/react`.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks! This is great.

PR #905, opened 2025-01-17
Fix: ensure resource routes return valid Response objects

hakimLyon:
Fixes resource route errors in React Router by ensuring each resource handler returns a valid response.
	•	Replaces the use of data with Response.json.
	•	Resolves the error: “Expected a Response to be returned from resource route handler”.

Cap


kentcdodds (approved):
Nice catch. Thanks!

PR #906, opened 2025-01-17
fix: layout shift caused by Dropdownmenu and make it to a separate component

rajeshg:
User drop down menu causes a horizontal layout shift (see https://github.com/shadcn-ui/ui/issues/977)

Also refactored UserDropdown to it's own component similar to SearchBar

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "sideEffects": false,
   "license": "MIT",
   "epic-stack": {
-    "head": "5c5328b38e442d78933d5be8fae21fe317296c29",
-    "date": "2025-01-17T19:16:26Z"
+    "head": "8468b0101fbee9d09f7ea426eaefec74583641b4",
+    "date": "2025-01-17T22:03:13Z"
   },
   "author": "Kent C. Dodds <me@kentcdodds.com> (https://kentcdodds.com/)",
   "type": "module",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -4,8 +4,8 @@
   "sideEffects": false,
   "license": "MIT",
   "epic-stack": {
-    "head": "5c5328b38e442d78933d5be8fae21fe317296c29",
-    "date": "2025-01-17T19:16:26Z"
+    "head": "8468b0101fbee9d09f7ea426eaefec74583641b4",
+    "date": "2025-01-17T20:19:33Z"
```
kentcdodds, line 8: Could you undo this change? Just leave out the `package.json` from the changes.

kentcdodds (changes requested):
Thanks for this! Just one thing. Also, can you verify that the actual user experience is unchanged?

package.json
rajeshg: Done

rajeshg:
> Thanks for this! Just one thing. Also, can you verify that the actual user experience is unchanged?

Yes. Looks good to me. Tested from Windows env

kentcdodds:
Thanks!

kentcdodds:
The `modal={false}` prop changed behavior and failed the playwright tests so I'm reverting the modal prop. I don't see any issues with layout shift with things...

PR #912, opened 2025-01-24
Add react router generated types folder to remix.init/gitignore

quangIle:
I think we miss this when migrating to React Router v7


kentcdodds (approved):
Thanks!

PR #913, opened 2025-01-28
fix uncaught exception

updbqn:
The check further down:

```typescript
if (userId && !user) {
		console.info('something weird happened')
		// something weird happened... The user is authenticated but we can't find
		// them in the database. Maybe they were deleted? Let's log them out.
		await logout({ request, redirectTo: '/' })
}
```

relies on the query returning null rather than throwing.

As an aside, I'm not sure exactly when `userId && !user` could happen. If the user was deleted, the onDelete cascade on Session means the previous getUserId call would already have redirected and removed the session cookie.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -78,7 +78,7 @@ export async function loader({ request }: Route.LoaderArgs) {
 	const user = userId
 		? await time(
 				() =>
-					prisma.user.findUniqueOrThrow({
+					prisma.user.findUnique({
 						select: {
 							id: true,
 							name: true,


jacobparis:
That's why it's prefaced with "something weird happened". What if the user was manually deleted in the database? what if there was an error in the onDelete and it never completed? This provides a graceful exit in cases like that

updbqn:
Sure, I get the reasoning. However, wouldn't the user have to disappear within milliseconds of the getUserId call for this to be an issue?

I think it's good to check our assumptions, but wouldn't it be better if we could do it once and be done with it? I would find that easier to read and reason about. I was hoping the aside might spark an idea for a cleaner approach.

Even further off-topic if it's interesting:

> What if the user was manually deleted in the database?

I would expect this to trigger the onDelete cascade?

> what if there was an error in the onDelete and it never completed?

I would expect the original delete and onDelete cascade to happen as one transaction and succeed or fail together?

Though I still agree that it's a good idea to try to handle weird states :)

PR #916, opened 2025-02-01
Fix: update alias configuration to resolve shadcn-ui/ui#6483

hakimLyon:
💫

### DIFF
--- a/components.json
+++ b/components.json
@@ -10,6 +10,7 @@
 	},
 	"aliases": {
 		"components": "#app/components",
-		"utils": "#app/utils/misc.tsx"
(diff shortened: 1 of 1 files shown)


components.json
```diff
@@ -10,6 +10,7 @@
 	},
 	"aliases": {
 		"components": "#app/components",
-		"utils": "#app/utils/misc.tsx"
+		"utils": "#app/utils/misc.tsx",
+		"ui": "#app/components/ui",
```
kentcdodds, line 14:
```suggestion
		"ui": "#app/components/ui"
```

kentcdodds (approved):
Thanks!

PR #917, opened 2025-02-03
sentry don't capture response

nichtsam:
I got this from sentry:
```
Error: Object captured as exception with keys: data, internal, status, statusText
```

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/components/error-boundary.tsx
+++ b/app/components/error-boundary.tsx
@@ -28,18 +28,23 @@ export function GeneralErrorBoundary({
 }) {
 	const error = useRouteError()
 	const params = useParams()
+	const isResponse = isRouteErrorResponse(error)
 
 	if (typeof document !== 'undefined') {
 		console.error(error)
 	}
 
 	useEffect(() => {
+		if (isResponse) {
+			return
+		}
+
 		captureException(error)
-	}, [error])
+	}, [error, isResponse])
 
 	return (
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks for this contribution. Can you verify that we are still getting Sentry errors from the server when we hit a 404 or a 500?

app/components/error-boundary.tsx
```diff
@@ -28,18 +28,23 @@ export function GeneralErrorBoundary({
...
+	const isResponse = isRouteErrorResponse(error)
 
 	if (typeof document !== 'undefined') {
 		console.error(error)
 	}
 
 	useEffect(() => {
+		if (isResponse) {
+			return
+		}
```
kentcdodds, line 40:
```suggestion
		if (isResponse) return
```

nichtsam:
> Thanks for this contribution. Can you verify that we are still getting Sentry errors from the server when we hit a 404 or a 500?

Thanks for reviewing! I tested it, and I believe we won’t get Sentry errors for 404s since they’re thrown manually, which I considered handled. However, 500 errors should still be captured by Sentry since they’re actual errors. Essentially, `throw new Response()` won’t be captured, but `throw new Error()` will.

kentcdodds:
Ok, I **think** 400s probably shouldn't be reported anyway so we're probably fine. Thanks!

PR #919, opened 2025-02-04
Validate password byte length

markuslewin:
Checks the byte length of the string to determine whether password is too long to `bcrypt.hash` properly.

## Test Plan

1. Try to create a new user with password `😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀a` (character length 37, byte length 73)
2. Password is too long


kentcdodds (approved):
Thank you! Great catch!

PR #922, opened 2025-02-06
fix: remove unused allowIndexing prop

danieltorrer:
Since `allowIndexing` now comes from an env variable, this prop is not used and can be safely removed

## Test Plan

Call to `<Document nonce={nonce} theme={theme} env={data?.ENV}>` does not complain


kentcdodds (approved):
Not sure how that happened, but thank you for the cleanup!

PR #923, opened 2025-02-08
refactor: simplified session query by removing redundant user join

AdityaKirad:
The current session query joins with the User table to get the userId, 
which is unnecessary since we can directly select userId from the Session table. 
Since we have onDelete: Cascade constraint, the database already guarantees 
referential integrity, making the join redundant.

- Removes unnecessary table join
- Improves query performance
- Relies on database constraints for data integrity


kentcdodds (approved):
Thanks!

PR #924, opened 2025-02-09
move away from express helmet

### LINKED ISSUE #920: CSP is being applied to non-HTML routes
Currently, Content Security Policy (CSP) is applied globally using the Express middleware helmet, which results in CSP headers being sent on all routes, including non-HTML responses.

It has two implications:
1. Unnecessary Overhead
CSP headers have no effect on non-HTML responses but still increase the response size.
2. Potential CORS Issues
Adding CSP headers to non-HTML responses may turn simple requests into non-simple ones, requiring CORS preflight checks.

nichtsam:
resolves #920 

## Issue:
Express Helmet applies security headers globally, which causes HTML-specific headers like Content-Security-Policy and X-Download-Options to be applied to non-HTML routes.

## Solution:
This PR replaces Express Helmet with a custom implementation that better distinguishes between general security headers, HTML-specific headers, and CORS-related configurations.

The new helmet function is inspired by [`helmet`](https://github.com/helmetjs/helmet) and [`http-helmet`](https://github.com/mcansh/http-helmet). 
There are two layers currently, one is the tool itself, and one is a wrapper around the tool with defined configurations.

### The new helmet tool

```ts
function helmet(options: HelmetOptions): Headers
export type HelmetOptions = {
	options: SecureOptions
	html?: boolean
	cors?: boolean
}
```
- The function simply returns the security headers with sensible defaults (referenced express helmet)
- The html specific options will only get applied if the `html` flag is enabled.
- `Cross-Origin-Resource-Policy` defaults to `'same-origin'`, but will default to `'cross-origin'` if the `cors` flag is enabled.
- I went through [Content-Security-Policy MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) and tried to make the interface organized, complete, correct, and auto-completing.

### The wrapper around helmet

This is the function to be used for handling the projects security headers.
I added function overloads to clearly indicate that nonce is only required for html content type.

The tool will probably be extracted into another package if this proof of concept is accepted, but I thought it'd be a good idea to have the interface reviewed here first.

## Test Plan

- Unit tests for each middleware to verify expected behavior in isolation.  
- Integration tests to ensure correct security headers in different scenarios (HTML vs. non-HTML, CORS enabled/disabled).  

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds:
Thank you so much for working on this! I like the direction here, and I know we have to get rid of Helmet. I'm really concerned about the amount of files added here and the complexity that everybody will take on. Feels like this should probably go into a library, and I would be happy to have it be a part of the Epic Web org if you're interested in helping me maintain it.

nichtsam:
> Thank you so much for working on this! I like the direction here, and I know we have to get rid of Helmet. I'm really concerned about the amount of files added here and the complexity that everybody will take on. Feels like this should probably go into a library, and I would be happy to have it be a part of the Epic Web org if you're interested in helping me maintain it.

Yes, definitely! I’m currently working on the library and preparing it for publishing on npm, but it will take some time.
Would it be okay to leave this PR as a draft and update it once the package is ready, or would you prefer to close it for now and open a new PR later?

kentcdodds:
I'll close it now and that way we don't lose track of where this is at while you work on the library. Thanks!

PR #926, opened 2025-02-11
fix: add missing name for query parameter redirectTo

### LINKED ISSUE #925: Malformed Signup Redirect URL
The link to the signup form generates an incorrect URL when `redirectTo` has a value because it omits the query parameter name `?redirectTo=`. This should be fixed to ensure valid redirection.

**Affected Line**

https://github.com/epicweb-dev/epic-stack/blob/7bb4bd7ab4749760661b7997b5dfda7cbce99f2c/app/routes/_auth%2B/login.tsx#L184

**Suggested Fix**

```diff
- to={redirectTo ? `/signup?${encodeURIComponent(redirectTo)}` : '/signup'}
+ to={redirectTo ? `/signup?redirectTo=${encodeURIComponent(redirectTo)}` : '/signup'}
```

roblesdotdev:
This PR fixes the malformed signup redirect URL by adding the missing redirectTo query parameter when a redirectTo value is present. This ensures that the redirect works correctly and includes the parameter name in the URL.

## Test Plan

This should fix #925

### DIFF
--- a/app/routes/_auth+/login.tsx
+++ b/app/routes/_auth+/login.tsx
@@ -181,7 +181,7 @@ export default function LoginPage({ actionData }: Route.ComponentProps) {
 							<Link
 								to={
 									redirectTo
-										? `/signup?${encodeURIComponent(redirectTo)}`
+										? `/signup?redirectTo=${encodeURIComponent(redirectTo)}`
 										: '/signup'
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

PR #928, opened 2025-02-13
Remove chalk

kinggoesgaming:
- **refactor: use styleText instead of chalk**
- **build(deps): remove chalk**

This is a rework of #798 

## Test Plan

## Screenshots

No change that requires diff of screenshots

closes: #792

### DIFF
--- a/app/utils/db.server.ts
+++ b/app/utils/db.server.ts
@@ -1,7 +1,7 @@
+import { styleText } from 'node:util'
 import { remember } from '@epic-web/remember'
 // Changed import due to issue: https://github.com/remix-run/react-router/pull/12644
 import { PrismaClient } from '@prisma/client/index.js'
-import chalk from 'chalk'
 
 export const prisma = remember('prisma', () => {
 	// NOTE: if you change anything in this function you'll need to restart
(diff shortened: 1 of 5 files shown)


app/utils/db.server.ts
```diff
@@ -29,7 +29,7 @@ export const prisma = remember('prisma', () => {
 						: e.duration < logThreshold * 1.4
 							? 'redBright'
 							: 'red'
-		const dur = chalk[color](`${e.duration}ms`)
+		const dur = styleText(color, `${e.duration}ms`)
```
kentcdodds, line 32: This is great except that the style text utility does not support the redBright color. So we'll have to come up with something else for that.

kentcdodds (changes requested):
Looks great, just one thing.

app/utils/db.server.ts
kinggoesgaming: According to the `ForegroundColors` type (which is the type for text colors), the following are allowed:

```ts
type ForegroundColors =
        | "black"
        | "blackBright"
        | "blue"
        | "blueBright"
        | "cyan"
        | "cyanBright"
        | "gray"
        | "green"
        | "greenBright"
        | "grey"
        | "magenta"
        | "magentaBright"
        | "red"
        | "redBright"
        | "white"
        | "whiteBright"
        | "yellow"
        | "yellowBright";
```

`redBright` appears to be allowed (and I am not seeing any type errors)

kentcdodds: Thanks for double checking on that!

Issue #931, opened 2025-02-15
Flaky tests in `tests/e2e/onboarding.test.ts`


nichtsam:
Tests inside `tests/e2e/onboarding.test.ts` seems to randomly fail, but succeed when ran individually.

potential Issues:
- Some tests check for pending state of status button, might fail if the pending state ends before it's tested.
- Some tests fail because click actions did not change states, as far as I seen only on radix checkboxes.
- Some tests fail because the msw mock github handlers aren't returning the inserted ghUser, causing mismatch and notfound.

I've recorded a short video to demonstrate the issue.
In the first run you can see that the user email doesn't match, in some other the user doesn't exist while it should have.

https://github.com/user-attachments/assets/626cc0b8-44e3-4607-ae6c-1f199e09b809

nichtsam:
Taken care of by #934

PR #934, opened 2025-02-18
improve playwright tests

nichtsam:
Related: #931

Flakiness reasons:
- Some tests check for pending states, but states might resolve before they get verified.
- `faker.internet.username()` can generate usernames that exceed the length limit defined in `SignupFormSchema`.
- Some tests interact with components that require JS before it's fully loaded (e.g., Radix Checkbox).

## Checklist

- [x] Tests updated
- [ ] Docs updated

### DIFF
--- a/tests/e2e/onboarding.test.ts
+++ b/tests/e2e/onboarding.test.ts
@@ -60,9 +60,6 @@ test('onboarding with link', async ({ page, getOnboardingData }) => {
 	await emailTextbox.fill(onboardingData.email)
 
 	await page.getByRole('button', { name: /submit/i }).click()
-	await expect(
-		page.getByRole('button', { name: /submit/i, disabled: true }),
-	).toBeVisible()
 	await expect(page.getByText(/check your email/i)).toBeVisible()
 
 	const email = await readEmail(onboardingData.email)
@@ -120,9 +117,6 @@ test('onboarding with a short code', async ({ page, getOnboardingData }) => {
 	await emailTextbox.fill(onboardingData.email)
 
 	await page.getByRole('button', { name: /submit/i }).click()
-	await expect(
-		page.getByRole('button', { name: /submit/i, disabled: true }),
-	).toBeVisible()
 	await expect(page.getByText(/check your email/i)).toBeVisible()
 
 	const email = await readEmail(onboardingData.email)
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
That looks good to me! Thanks!

nichtsam:
> That looks good to me! Thanks!

Hey! I'm working on this partially as a showcase of issues, that's why I wrote the `reliableAction` stuff, not sure if it's a good solution.
I don't think I solve every flakiness yet, but I think I reduced it a lot by these changes, at least on my device 😅

kentcdodds:
I don't really want to go with the reliable action approach. I would like to dig a little deeper.

kentcdodds (approved):
These are good improvements. Thank you a lot!

andrecasal:
I've identified two situations where tests fail:
- A `page.goto()` does a full page reload, which requires hydration. If we interact with *any* form components before hydration occurs, the interaction fails.
- When we click a link to load another page, individual JS files are loaded for each component the client hasn't had a change to request yet. This doesn't require hydration, but it does require the loaded JS to run before we interact with the components. However, there's an exception: if the component required for the linked page has been loaded at some point before clicking the link, the client has already cached that JS and it just needs to run—which is considerable faster in comparisong to network requests, but not necessarily a guarantee that the test will pass.

Manually tracking `page.goto()` and which components have been loaded in previous pages isn't a reliable or transparent technique.

So my question is: wouldn't progressively enhanced components, i.e. components that work correctly before JS has loaded/with JS disabled, and also respect whatever state the component had upon hydration/JS run, be a better approach? Am I missing anything? What's your take on this @kentcdodds?

These test failures could just as easily occur in slow networks too.

kentcdodds:
Thanks for bringing this up.

You can't always rely on progressive enhancement, and on top of that, most users are going to have JavaScript actually running. So it is better to simulate the actual environment that users are going to experience rather than rely on progressive enhancement, which most users will not be relying on.

kentcdodds:
I think it really depends on what your priorities are and your user base. The vast majority of users are using your site with JavaScript enabled. If the progressive enhancement case matters for you, then having specific tests for that makes sense. I don't think that it's significant enough for this project for us to do that by default though, so I think we'll keep things the way they are.

PR #935, opened 2025-02-19
Add openimg for image optimization

andrelandgraf:
Hey! Currently, the Epic Stack doesn't have a solution for image optimization and I would like to bring this up to discussion. I have been working on [openimg](https://github.com/andrelandgraf/openimg), a small image optimization toolkit, as I have always wanted a small simple solution particularly for my Remix/React Router apps. I think it's simple enough to be included and provide value right out of the gates!

## openimg

The goal of [openimg](https://github.com/andrelandgraf/openimg) is to be easy to use (inspired by Vercel's image optimization services) but also highly scalable and configurable, so you can reconfigure it (or replace it) as your app grows.

Features:

- openimg/node, openimg/bun: Configurable HTTP request handler for optimizing images and other utils (like base64 placeholder generation)
- openimg/react: React component (and context provider) to query images from the optimization endpoint using modern image/picture attributes
- openimg/vite: A vite plugin (optional) to generate Img React components with base64 placeholders for Vite assets postfixed with`?react`

You can start simple by adding a new resource route for the image optimization and then query images from the endpoint (see this PR). Next, you can add the `Img` React component to upgrade each image on your site incrementally (also already included in this PR). If you grow to have a lot of images, you can easily break out the optimization endpoint into a standalone server ("for scale") and only change a few lines of code to redirect all optimization requests to the remote server. Further, you can both optimize locally hosted image and remote images (e.g., from an S3 bucket) and use file system caching (default) or just utilize a CDN with HTTP caching headers, this should make openimg configurable enough to work across different environments (serverless, long-running server).

I have been using openimg in production on [allthingsweb.dev](https://allthingsweb.dev) where the images are stored on S3 and optimized and cached on a Fly-hosted React Router app. Very happy with how it's been working!

## Epic Stack

I understand the Epic Stack attempts to reduce services, including the self-managed variety, as per the [image decision doc](https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/018-images.md). Still, I think image optimization is important enough to build a solid starting point that scales with your app for a truly epic experience!

The main goal of image optimization is to reduce the total bytes a user downloads. Smaller images can make a big difference, even on sites with only a few images. For example, on Kody's note details page, we can shrink the image load from **1112kb** to **46.5kb** by resizing and converting images to AVIF on the fly.

## Before & After

Before, 456kb + 348kb + 308kb:

Screenshot 2025-02-19 at 9 37 31 AM

After, 19kb + 13.1kb + 24.4kb:

Screenshot 2025-02-19 at 9 38 19 AM

Note that the initial resizing work does increase the response times for the first load (~from 10ms locally to ~200-300ms):

Screenshot 2025-02-19 at 9 37 58 AM

However, this is a trade-off for the first request of a given image (w, h, format) only. After the initial resizing and transformation, we can cache the resized image on the filesystem (default) and/or via CDN & HTTP caching. Using file system caching, we can then even slightly reduce the load times because we need to load less bytes (see second screenshot).

## HTML picture element

The `Img` component from `openimg/react` uses the picture element to query for the smallest possible image with modern formats, but can also fallback to the img element based on browser capabilities. The goal is to provide the least amount of information (just width & height) to create a good-enough image optimization approach. I personally don't want to deal with `srcset` and `sizes` queries for every image that I add to my site.... but of course you can pass in and override all attributes.

The default looks something like this:

```html
<picture>
    <source width="516" height="516" srcset="/resources/images?src=%2Fresources%2Fnote-images%2Fcm7c73fm400146dr24wl3fz3k&amp;w=516&amp;h=516&amp;format=avif" sizes="516px" type="image/avif" />
    <source width="516" height="516" srcset="/resources/images?src=%2Fresources%2Fnote-images%2Fcm7c73fm400146dr24wl3fz3k&amp;w=516&amp;h=516&amp;format=webp" sizes="516px" type="image/webp" />
    <img
        width="516"
        height="516"
        loading="lazy"
        decoding="async"
        fetchpriority="low"
        alt="an adorable koala cartoon illustration"
        class="h-32 w-32 rounded-lg object-cover"
        src="/resources/images?src=%2Fresources%2Fnote-images%2Fcm7c73fm400146dr24wl3fz3k&amp;w=516&amp;h=516"
    />
</picture>
```

Note you can add a `isAboveFold` prop to update the `loading`, `decoding` and `fetchpriority` attributes to prioritize an image!

With the changes in this PR, I was able to  address the "use modern formats" lighthouse recommendation:

Screenshot 2025-02-19 at 10 28 20 AM

I believe these changes will help users get more mileage out of Epic Stack, but I would love to hear your thoughts on this topic!

## Test Plan

TBD

## Checklist

TBD

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

No visual changes, just smaller images in modern formats!


.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 .DS_store
 
+/data
```
andrelandgraf, line 4: Only needed if the default filesystem caching is enabled. Default caching destination should work out of the box with the current Fly setup!

app/components/user-dropdown.tsx
```diff
@@ -25,10 +26,12 @@ export function UserDropdown() {
 						onClick={(e) => e.preventDefault()}
 						className="flex items-center gap-2"
 					>
-						<img
+						<Img
 							className="h-8 w-8 rounded-full object-cover"
 							alt={user.name ?? user.username}
 							src={getUserImgSrc(user.image?.id)}
+							width={416}
+							height={416}
```
andrelandgraf, line 34: I was pretty conservative with these max-width values to have all images look and feel high quality and not pixelated. Note that the CSS width and height will override the HTML attributes, so there are no visual changes. The width & height attributes are still important for the browser to compute the aspect ratio and for openimg to request a good-enough resized image.

.gitignore
kentcdodds: So if I understand it right, you're putting the caching destination as the current working directory`/data`. And I worry that that is going to get blown away on every redeploy of the app. I think that we might want to configure that to go somewhere that would be persisted between deploys.

app/components/user-dropdown.tsx
kentcdodds: Honestly, that's probably what I would have done too. 👍

kentcdodds: Actually, I think that this is going to be 256x256, and I think that we will get plenty of high-quality images by setting it to `256`.

app/routes/users+/$username_+/__note-editor.tsx
```diff
@@ -188,10 +189,12 @@ function ImageChooser({ meta }: { meta: FieldMetadata<ImageFieldset> }) {
 						>
 							{previewImage ? (
 								<div className="relative">
-									<img
+									<Img
 										src={previewImage}
 										alt={altText ?? ''}
 										className="h-32 w-32 rounded-lg object-cover"
+										width={516}
+										height={516}
```
kentcdodds, line 197:
```suggestion
										width={512}
										height={512}
```

package.json
```diff
@@ -105,6 +106,7 @@
     "remix-auth-github": "^1.7.0",
     "remix-utils": "^8.1.0",
     "set-cookie-parser": "^2.7.1",
+    "sharp": "^0.33.5",
```
kentcdodds, line 109: Why is this a pure dependency of OpenImg and not just a built-in dependency? Why do I need to install it myself?

kentcdodds (changes requested):
I love this a lot, just a couple of things.

.gitignore
andrelandgraf: Good points. The caching strategy depends on how the Epic Stack is deployed. By default, optimized images are cached to `./data/images` but this can be configured via the `cacheFolder` argument: `string` | `no_cache` ([API reference](https://github.com/andrelandgraf/openimg/blob/main/docs/api/node.md#config-options)). Since the Epic Stack is hosted on Fly with a mounted volume, I assumed the `/data` folder would persist between deployments:

```toml
[mounts]
source = "data"
destination = "/data"
```

I guess every machine has its own volume/image cache in the current setup, but I believe they persist?

I guess our options are:

- Quasi-persistent cache to the filesystem at `/data` if Fly persists the volumes per machine in the current setup
- If we have a CDN like Cloudflare running in front of the machines, then HTTP caching is probably sufficient with `cacheFolder: "no_cache"`.
-  Synchronized caching across all machines via SQLite as a cache location. I think that would require the least changes to the Epic Stack, but would require some code changes to openimg. Happy to add that via a `customCache` config option.

Lmk what you would prefer!

app/routes/resources+/images.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { getDomainUrl } from '#app/utils/misc.tsx'
+
+function isUrl(src: string) {
+	try {
+		new URL(src)
+		return true
+	} catch {
+		return false
+	}
+}
```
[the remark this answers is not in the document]
andrelandgraf: TIL! 🔥

package.json
andrelandgraf: This decision was made based on a gut feeling. Sharp uses [libvips](https://github.com/libvips/libvips) under the hood, and bundling binaries can get tricky. I figured it was better to let you control how to bundle sharp and choose the version you need. I believe Next.js takes a similar approach by requiring you to install sharp yourself (see [here](https://nextjs.org/docs/messages/sharp-missing-in-production)). But happy to change my mind if you think it's better otherwise!

kentcdodds: Sounds good to me!

andrelandgraf:
Hey Kent, thanks for looking into this PR so quickly and for the great discussions! I added a first draft for the decision doc and documentation for image optimization (including the existing image logic).

.gitignore
kentcdodds: If you're sticking things in `./data/images`, that's going to be relative to the repo, not the mounting destination (right now, I think this is going into `/data/myapp/data/images`. So you'll want to make sure that it's not relative to the repo, but rather the specific mount destination, which will make things persisted. 

We do not have a CDN like Cloudflare running in front of the machines. Our server serves up all the static assets directly. 

I think the most correct option here would be to use Tigris for this. I was planning on doing Tigris soon anyway, so I'll get that set up, and then we can get this working with Tigris as our storage location. Do you think that would work?

kentcdodds (commented):
Thanks for giving me the motivation I needed to get Tigris working for the Epic Stack 😅

kentcdodds:
Thank you so much for working on this. It really got me going on Tigress hosting, and I'm excited to integrate this with what I have coming very soon. I'm sorry, but you're going to have a couple of merge conflicts. I hope it's not very significant, but when we're all done, this is going to be so epic. If you're curious, here's the live stream: https://www.youtube.com/watch?v=flZCS8KthzU

I got so close to being done! You can also check the dev branch which has all the work. Pretty sure it's ready.

kentcdodds:
https://github.com/epicweb-dev/epic-stack/compare/dev

kentcdodds:
I have successfully finished the migration to Tigris for storing our images. If you could update your PR to handle any changes, that would be great. Hopefully, it's pretty minimal. This will be much better. Thank you!

.gitignore
andrelandgraf: You are so right. 🤦  Well, thanks for [catching this](https://github.com/allthingsweb-dev/allthingsweb/commit/757b36d88902828eb4b1ebb514518860a6e208b7)! Fixed! For local dev, I think we still want it in the repo but for prod, we definitely want it at the proper volume mounting location.

app/routes/resources+/images.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { type Route } from './+types/images.ts'
+import { getDomainUrl } from '#app/utils/misc.tsx'
+
+export async function loader({ request }: Route.LoaderArgs) {
+	const domain = getDomainUrl(request)
+	const headers = new Headers()
+	headers.set('Cache-Control', 'public, max-age=31536000, immutable')
+	return getImgResponse(request, {
+		headers,
+		allowlistedOrigins: [domain, process.env.AWS_ENDPOINT_URL_S3],
```
andrelandgraf, line 11: We don't actually fetch an image from Tigris directly just yet but I already added the `process.env.AWS_ENDPOINT_URL_S3` in the allowlist.

andrelandgraf:
> I have successfully finished the migration to Tigris for storing our images. If you could update your PR to handle any changes, that would be great. Hopefully, it's pretty minimal. This will be much better. Thank you!

Awesome! I didn't actually need to make any code changes. This is because we are consuming the existing resource user and note endpoints in `+resources/images`. This is actually an indirection that we should probably remove? If the client-side user and note objects would include the signed image URLs already, then we could query the images via: `/resources/images?${user.profileImageUrl}`. From there, the images endpoint could fetch the image from Tigris directly instead of fetching the image from the user and note endpoints. I think this is good to merge, but it may be worth rearchitecting in the future. Lmk what you think!

kentcdodds:
I think I would like the user and note images to use openimg as well. What do you think?

andrelandgraf:
They are! 100%, we want to build a solution that covers all images. The way note & user images are currently optimized is via this flow:

- `Img` component queries from `/resources/images` with `src` search parameters like `/resources/users/123`.
- `/resources/images`: If the requested image is in the cache, retrieve it from there.
- `/resources/images`: Else, fetch the original image from `/resources/users/123`, optimize it, cache it, and serve it.

This is the configuration in `/+resources/images` for this:

```typescript
if (params.src.startsWith('/resources')) {
  // Fetch image from resource endpoint
  return {
    type: 'fetch',
    url: domain + params.src,
  }
}
```

The indirection I wanted to highlight is making a fetch request to another endpoint on the same server. It's fine as it only happens on cache missed (first request), but we could still optimize it. The easiest way would be to remove these resource routes altogether and always return all signed URLs to the client so that the client can immediately query `/resources/images` with the src being the signed S3 bucket URL. But I don't think this needs to be part of this PR.

app/routes/resources+/images.tsx
```diff
@@ -0,0 +1,91 @@
...
+		select: { objectKey: true },
+	})
+	invariantResponse(userImage, 'User image not found', { status: 404 })
+
+	const { url, headers } = getSignedGetRequestInfo(userImage.objectKey)
+	return {
+		type: 'fetch',
+		url,
+		headers,
+	}
```
kentcdodds, line 75: I need the fetch request you make to include these headers.

I also need `getImgSource` to support returning a promise.

kentcdodds (changes requested):
Ok, so I never like it when a server makes requests to itself, so I've updated this PR to not do that, but we need some changes in openimg for this to work. Could you look into that?

kentcdodds:
We're very close! This is a big improvement!

app/routes/resources+/images.tsx
andrelandgraf: Available via `openimg` v0.2.0! 🎉

app/utils/misc.tsx
```diff
@@ -6,11 +6,11 @@ import { extendTailwindMerge } from 'tailwind-merge'
 import { extendedTheme } from './extended-theme.ts'
 
 export function getUserImgSrc(imageId?: string | null) {
-	return imageId ? `/resources/user-images/${imageId}` : '/img/user.png'
+	return imageId ? `/resources/images?userImageId=${imageId}` : '/img/user.png'
```
andrelandgraf, line 9: This works. It generates this HTML img element src attribute: `/resources/images?src=/resources/images?userImgId=123`. 

- Path `/resources/images` comes from the `OpenImageContextProvider`'s `optimzierEndpoint` prop
- Src query comes from `getUserImgSrc`

Note that we don't need to add `/resources/images` here because we only use the `userImageId` search parameter value in `getImgSources`. Maybe you decided to do this to be more explicit, but just wanted to point it out in case this is an oversight. I also can't think of a better path. 😅

andrelandgraf (commented):
Love the changes you made. It's much more straightforward now. Added one comment only!

app/utils/misc.tsx
kentcdodds: Yeah, that’s weird. Any way to make it not that way?

andrelandgraf: The problem is that we have to map from `imageId` to `imageSrc` to signed `imageSrc`. Taking a step back, I think we want to get to the point that adding an image in the UI looks like this:

```tsx
<Img src={user.imgSrc} />
```

I like this a lot because `openimg` becomes just an implementation detail. You can replace `Img` with `img` and it still works (a bit like Remix's `Form` component!) because the `src` is a valid image source. If we want to get to this point, we need to map the `imageId` to `imageSrc` in the loader functions or store the `imageSrc` in the database instead of the `imageId`. What do you think?

kentcdodds: I prefer the function call. I just want to think about making the path not so odd. I would prefer to make it be like this: `/resources/images/$objectKey`. This way when people remove the notes they don’t have to worry about special handling for notes. So I think we’ll need to do a lookup for the object ID for the UserImage and NoteImages.

andrelandgraf: Okay! The goal is definitely to have `openimg` be configurable in whatever way works best for you:

- The `OpenImageContextProvider` supports a `getSrc` function where you can transform the `Img` props to the img `src` string
- The `getImgResponse` function also has a `getImgParams` function that is called before `getImgSource` to map the `request.url` to `w`, `h`, `fit`, `format`, and `src`.
- We could also refactor `+/resouces/images.tsx` to `+/resources/images/$.tsx` to accept slugs.

So we should have all the flexibility to design a solution that you prefer. I have something ready in a bit!

andrelandgraf:
Hey @kentcdodds, I went through a few iterations, but pretty happy now with the outcome. Lmk what you think!

app/utils/misc.tsx
```diff
@@ -2,15 +2,40 @@ import { clsx, type ClassValue } from 'clsx'
...
+		if (fit) {
+			searchParams.set('fit', fit)
+		}
+		if (format) {
+			searchParams.set('format', format)
+		}
+		return `${endpoint}?${searchParams.toString()}`
+	}
+	return defaultGetSrc({ height, optimizerEndpoint, src, width, fit, format })
 }
```
kentcdodds, line 39: Any reason this isn't just the default behavior?

andrelandgraf: I would like that adding openimg is just as easy as replacing `img` with `Img`: ``  to ``. That's what I built my logic around. But here, we want our source to specify the optimizer endpoint directly.

kentcdodds: That makes sense. I think I will add a comment about that.

kentcdodds (approved):
Thanks so much, this is way cool!

PR #944, opened 2025-02-21
Chore/update remix auth GitHub

### LINKED ISSUE #932: Update the remix-auth packages
When trying to implement then newest versions of remix-auth-microsoft, it seems incompatible with the current remix-auth version. I tried upgrading and managed to get it working, but did not change the mocking functionality and got ReferenceError: Missing state on cookie. It seem state management via cookies is different with the newer versions. How can we make mocking work when going from this:

"remix-auth": "^3.7.0"
"remix-auth-github": "^1.7.0"

to this?

"remix-auth": "^4.1.0",
"remix-auth-github": "^3.0.2",

rperon:
This PR update remix-auth and remix-auth-github to the latest version, to close #932 .

The main chalenge in this PR was to update the mock implementation done by remix-auth-github.

## Test Plan

This needs to be deployed and tested with Github.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


rperon:
Sorry, there is one test failing, for some reason we don't receive the throw from the loader...

app/utils/connections.server.ts
```diff
@@ -1,21 +1,9 @@
...
-		name: 'en_connection',
-		sameSite: 'lax', // CSRF protection is advised if changing to 'none'
-		path: '/',
-		httpOnly: true,
-		maxAge: 60 * 10, // 10 minutes
-		secrets: process.env.SESSION_SECRET.split(','),
-		secure: process.env.NODE_ENV === 'production',
-	},
-})
-
```
kentcdodds, line 18: I'm confused why changing this was necessary because it looks like we've now lost the ability to display errors as the user goes between different pages in the OAuth flow. And I don't think that this change was necessary for upgrading to Remix auth. Can you help me understand why this change was made?

kentcdodds (changes requested):
I'm very much looking forward to this change. Thank you so much for working on it. I just have a couple of questions and we'll want to make sure the tests all pass as well.

app/utils/connections.server.ts
sergiodxa: This is no needed anymore since Remix Auth doesn't receive a session storage instance anymore, instead the strategy uses [`@mjackson/headers`](https://www.npmjs.com/@mjackson/headers) to save the data it needs in a cookie during the auth process.

Then after the user is back in the app from GitHub it you need to save the data in your session storage yourself, AFAIK you already did this anyway so there's no change in that part.

sergiodxa: Also related to error handling, now the strategies always throw an error, so you handle them with a try/catch

kentcdodds: Great to know! Thanks!

app/routes/_auth+/auth.$provider.callback.test.ts
```diff
@@ -35,7 +35,7 @@ test('when auth fails, send the user to login with a toast', async () => {
 	consoleError.mockImplementation(() => {})
 	server.use(
 		http.post('https://github.com/login/oauth/access_token', async () => {
-			return new Response('error', { status: 400 })
+			return new Response(null, { status: 400 })
```
rperon, line 38: I had to change from "error" to null because the test was in timeout, stuck in the `await response.body.cancel()` call inside the `arctic` library. Cf. screenshots.
Screenshot 2025-02-22 at 16 03 52

rperon:
Thanks for your feedback. Grad to help :)
I fixed the test, do you have another feedback for me ? 

Of course this still has to be tested on a deployed server ;)

kentcdodds (approved):
Thank you so much! I’m going to merge and pray it works when deployed. If not I’ll fix it quick 😅

this is something I’ve wanted done in a while so I really appreciate your help!

kentcdodds:
Well, it took a couple tries, but it's working in production now 😅

rperon:
Well done 👍

kentcdodds:
Seriously, thank you! I had tried upgrading remix auth once or twice and I just never dug in deep enough to figure out what changes needed to happen so it's been in the back of my mind for a while. I'm so glad to have this now!

PR #945, opened 2025-02-21
Migrate from storing images in SQLite to Tigris

kentcdodds:
The reasoning is all in the decision doc that's in this set of code changes.

## Test Plan

I verified that this works in staging as well as local development, even if you don't have the environment variables set. If you have mocking enabled, which is on by default in development, everything works locally with those mocks. Existing tests passes.

## Checklist

- [x] Tests updated
- [x] Docs updated

## Screenshots

No visual changes.


kentcdodds (commented):
Great job Kent! 😆

PR #952, opened 2025-02-25
Feat/small tailwindv4 improvements

DennisKraaijeveld:
Some small adjustments made to the .css file and added the v4 components straight from his v4 registry. I am not deep into the Epic Stack (styling) anymore so I might miss quite some things so consider this as another checkpoint.


kentcdodds:
Merged #968. Thanks!

Issue #953, opened 2025-02-27
JSX Key Bug


ScottDalessandro:
On `/users/kody/notes/new`, on the first load there is a jsx key props bug

![Image](https://github.com/user-attachments/assets/89b39c87-a7a5-4b20-99e7-88df3819f1df)

It also happens on edits as well (e.g. `/users/kody/notes/d27a197e/edit`).

![Image](https://github.com/user-attachments/assets/b5e1d094-7abe-49ef-b46e-e7afb0dc8d6d)

I will try to debug this soon.

ScottDalessandro:
Note: This happens when I first start in dev mode and visit the new notes route. If I click a few notes and then click new note, it doesn't appear to have a jsx bug? Just noting this behavior, pressed for time atm but will look into this soon.

braginteractive:
This is due to how conform passes the key with the inputprops. I ended up editing all my form elements. 

For example:

export function Field({
	labelProps,
	inputProps,
	errors,
	className,
}: {
	labelProps: React.LabelHTMLAttributes<HTMLLabelElement>
	inputProps: React.InputHTMLAttributes<HTMLInputElement> & {
		key?: string
	}
	errors?: ListOfErrors
	className?: string
}) {
	const fallbackId = useId()
	const id = inputProps.id ?? fallbackId
	const errorId = errors?.length ? `${id}-error` : undefined

	const { key, ...restInputProps } = inputProps

	return (
		<div className={className}>
			<Label htmlFor={id} {...labelProps} />
			<Input
				id={id}
				key={key}
				aria-invalid={errorId ? true : undefined}
				aria-describedby={errorId}
				{...restInputProps}
			/>
			<div className="min-h-[32px] px-4 pb-3 pt-1">
				{errorId ? <ErrorList id={errorId} errors={errors} /> : null}
			</div>
		</div>
	)
}

Seems conform has tried to solve for this is the past but eneded up rolling back in the latest release. Not sure if conform is stil being maintained, so it might be time to consider switching this out. Something like Tanstack Form or React Hook Form depending on the direction shadcn takes? 

@kentcdodds thoughts on the direction of form and fields? Shadcn hinted at a form refresh..

PR #954, opened 2025-02-28
add explicit key prop to JSX element

ScottDalessandro:
[Please check Issue](https://github.com/epicweb-dev/epic-stack/issues/953)

## Test Plan

On a new epic stack instance, visit and click the new note page visiting `/users/:username/notes/new` and the browser console should be empty with errors:

image

## Checklist

- [ ] Tests updated: No
- [ ] Docs updated: No

## Screenshots

Before (when creating a new note)

image

After:

image

### DIFF
--- a/app/routes/users+/$username_+/__note-editor.tsx
+++ b/app/routes/users+/$username_+/__note-editor.tsx
@@ -184,7 +184,7 @@ function ImageChooser({
 	const [altText, setAltText] = useState(fields.altText.initialValue ?? '')
 
 	return (
-		<fieldset {...getFieldsetProps(meta)}>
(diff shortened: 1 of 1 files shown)


updbqn (commented):
I think conform expects us to use the key from field for everything to work correctly.

See e.g. https://github.com/edmundhung/conform/issues/600#issuecomment-2074577745

ScottDalessandro:
Nice find, thanks! So the option here is to ignore them. I'll try to make a demo production version of this. I'd presume React would suppress these errors in production and we'll only have to deal with them in dev environments.

zwhitchcox:
This should be fixed by https://github.com/edmundhung/conform/pull/837, but will need to upgrade to `1.4.0

PR #958, opened 2025-03-06
prevent use of common password using the haveibeenpwned password api

AdityaKirad:
added checks in the schemas of onboarding and reset-password page to prevent the use of common password using the HaveIBeenPwned Password API

### DIFF
--- a/app/utils/auth.server.ts
+++ b/app/utils/auth.server.ts
@@ -1,3 +1,4 @@
+import crypto from 'node:crypto'
 import { type Connection, type Password, type User } from '@prisma/client'
 import bcrypt from 'bcryptjs'
 import { redirect } from 'react-router'
@@ -254,4 +255,34 @@ export async function verifyUserPassword(
 	}
 
 	return { id: userWithPassword.id }
+}
+
+export async function checkCommonPassword(password: string) {
+	const hash = crypto
+		.createHash('sha1')
+		.update(password, 'utf8')
+		.digest('hex')
+		.toUpperCase()
+
+	const [prefix, suffix] = [hash.slice(0, 5), hash.slice(5)]
+
+	const controller = new AbortController()
+
+	try {
+		const timeoutId = setTimeout(() => controller.abort(), 1000)
+
+		const res = await fetch(`https://api.pwnedpasswords.com/range/${prefix}`)
+
+		clearTimeout(timeoutId)
+
+		if (!res.ok) false
+
+		const data = await res.text()
+		return data.split('/\r?\n/').some((line) => line.includes(suffix))
+	} catch (error) {
+		if (error instanceof DOMException && error.name === 'AbortError') {
+			console.warn('Password check timed out')
+		}
+		return false
+	}
 }
--- a/app/utils/auth.server.ts
+++ b/app/utils/auth.server.ts
@@ -283,6 +283,7 @@ export async function checkCommonPassword(password: string) {
 		if (error instanceof DOMException && error.name === 'AbortError') {
 			console.warn('Password check timed out')
 		}
+		console.warn('unknow error during password check', error)
 		return false
 	}
 }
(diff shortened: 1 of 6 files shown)


kentcdodds (changes requested):
Thanks for this! I think it's a good idea. Just a couple things:
1. Let's reduce the duplication by putting this logic in `auth.server.ts` as a util and using that util.
2. Let's add a MSW mock in the `test/mocks` directory so we don't depend on it for development and testing
3. Let's add a timeout to the request so if it takes more than 1 second we skip the check
4. If the request fails, let's skip the check

I don't want a user to not be able to set their password just because a third party service is down or slow or unreliable. And all third party requests we make should be mocked out.

Thanks for this! I think it's great!

AdityaKirad:
will make the necessary changes asap

app/utils/auth.server.ts
```diff
@@ -255,3 +256,33 @@ export async function verifyUserPassword(
...
+		const res = await fetch(`https://api.pwnedpasswords.com/range/${prefix}`)
+
+		clearTimeout(timeoutId)
+
+		if (!res.ok) false
+
+		const data = await res.text()
+		return data.split('/\r?\n/').some((line) => line.includes(suffix))
+	} catch (error) {
+		if (error instanceof DOMException && error.name === 'AbortError') {
```
kentcdodds, line 283: In Node.js the error thrown is an `AbortError` not a `DOMException`. Could you update this? Could probably just do this:

```suggestion
		if (error instanceof Error && error.name === 'AbortError') {
```

tests/mocks/common-password.ts
```diff
@@ -0,0 +1,8 @@
+import { http, HttpResponse } from 'msw'
```
kentcdodds, line 1: Let's follow the convention of other services. Call the file: `pwnedpasswords.ts` and export an array of handlers called `handlers`.

kentcdodds (changes requested):
Thanks! Just two things :)

kentcdodds (changes requested):

app/utils/auth.server.ts
```diff
@@ -255,3 +256,33 @@ export async function verifyUserPassword(
...
+
+		if (!res.ok) false
+
+		const data = await res.text()
+		return data.split('/\r?\n/').some((line) => line.includes(suffix))
+	} catch (error) {
+		if (error instanceof Error && error.name === 'AbortError') {
+			console.warn('Password check timed out')
+		}
+		return false
```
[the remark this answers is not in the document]
AdityaKirad: Yeah completely forgot about it

app/utils/auth.server.ts
AdityaKirad: actually, it was copilot mistake, and I didn't paid attention to it sorry for that

app/utils/auth.server.ts
```diff
@@ -283,6 +283,7 @@ export async function checkCommonPassword(password: string) {
 		if (error instanceof Error && error.name === 'AbortError') {
 			console.warn('Password check timed out')
 		}
+		console.warn('unknow error during password check', error)
```
kentcdodds, line 286:
```suggestion
		} else {
			console.warn('unknow error during password check', error)
		}
```

kentcdodds (changes requested):
No need to give two warnings.

tests/mocks/index.ts
```diff
@@ -1,5 +1,6 @@
 import closeWithGrace from 'close-with-grace'
 import { setupServer } from 'msw/node'
+import { pwnedPasswordApiHandlers } from './common-password.ts'
```
kentcdodds, line 3: Let's rename the common-password file to `pwnedpasswords.ts`

kentcdodds (approved):
Thank you! This is a nifty feature 👏

PR #960, opened 2025-03-07
remove samplerate, as it is superseded by tracesSampler

zwhitchcox:
`tracesSampler` overrides the flat sample rate:

https://docs.sentry.io/platforms/javascript/configuration/sampling/#precedence

Screenshot 2025-03-07 at 9 04 57 AM

[We already check if environment is in production with the SENTRY_ENABLED variable](https://github.com/epicweb-dev/epic-stack/blob/main/server/index.ts#L18-L22), so the production check is redundant


server/utils/monitoring.ts
```diff
@@ -10,7 +10,6 @@ export function init() {
 	Sentry.init({
 		dsn: process.env.SENTRY_DSN,
 		environment: process.env.NODE_ENV,
-		tracesSampleRate: process.env.NODE_ENV === 'production' ? 1 : 0,
```
kentcdodds, line 13: I think it would make sense to move this prod check to the `tracesSampler` function

kentcdodds (changes requested):
Thanks! TIL!

kentcdodds (approved):
Thanks!

PR #963, opened 2025-03-13
fix: clearing profile name

updbqn:
Currently submitting an empty name on `/settings/profile` does not get persisted to the database. 

This happens due to empty string being coerced to undefined by conform and prisma treating undefined values  the same as the value not being provided at all.

Not sure if  there is some establish pattern for this. I'm guessing even people using zod and prisma without conform would like to avoid mixing "" and null for optional data.


kentcdodds (approved):
Looks good to me. Thanks!

PR #964, opened 2025-03-15
Improve common password checking

ryanwoodcox:
This is a follow up to #958 

- rename checkCommonPassword to checkIsCommonPassword for clarity
- use AbortSignal.timeout instead of manual timeout handling
- fix regex pattern for line splitting
- properly parse HIBP API response format
- improve error handling and messages
- add comprehensive tests for all scenarios
- extract getPasswordHashParts as a separate function (to be leveraged by the tests)

## Test Plan

- Run `npm run test` and see all tests including the 5 new tests pass.

## Checklist

- [x] Tests updated
- [ ] Docs updated


kentcdodds (changes requested):
Great improvements. Thank you!

PR #968, opened 2025-03-24
upgrade to tailwind v4 and some minor tweaks

AdityaKirad:
upgrade to tailwind v4 moved all the presets and extend theme config to `tailwind.css` updated the Shadcn components removed forwardRef and passed the ref directly to the components and made some minor tweaks to css of components like replacing `w-4 h-4` with `size-4` or `left-0 right-0` with `inset-x-0`


kentcdodds:
Thank you so much for this! Could you go through and make sure there are no unintentional UI differences? I noticed a few just going through things a bit myself:

image

image

image

AdityaKirad:
will make the necessary changes

kentcdodds (approved):
This is fantastic. Thank you so much for putting this work in 🙌

PR #974, opened 2025-03-30
docs(routing): Update the routes command

payapula:
`npx remix routes` returns error: "could not determine executable to run"

So, updated it to `npx react-router routes`

## Checklist

- [ ] Tests updated
- [x] Docs updated


kentcdodds (approved):
Thanks!

PR #977, opened 2025-03-31
updated getting-started.md initial command wording

### LINKED ISSUE #973: [Docs] Mention the node version to use in Getting Started Docs
I followed the [getting-started doc](https://github.com/epicweb-dev/epic-stack/blob/main/docs/getting-started.md) to start a new project. 

After running the `npm run dev` command, I ran into this issue

```bash
Error: No such built-in module: node:sqlite
    at ModuleLoader.builtinStrategy (node:internal/modules/esm/translators:444:11)
    at callTranslator (node:internal/modules/esm/loader:428:14)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:434:30)
    at link (node:internal/modules/esm/module_job:87:21)
```

Image

Then after researching for a bit I was using the node version v20. When I updated to latest version of node `v22.13.1` the error goes away. 

I noticed the [node decisions document](https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/021-node-version.md) and found that it mentions we always need to use latest LTS version of node in epic-stack. Which I guess must be added in the getting started documentation as well.

damiensedgwick:
Updated `getting-started.md` to include an additional instruction for using the current LTS version of Node.js to avoid any issues. 

Fixes #973 

## Test Plan

No testing needed, just a small documentation update.

## Checklist

- [ ] Tests updated
- [X] Docs updated

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,7 +2,7 @@
 
 The Epic Stack is a [Remix Stack](https://remix.run/stacks). To start your Epic
 Stack, run the following [`npx`](https://docs.npmjs.com/cli/v9/commands/npx)
-command:
+command using the current LTS version of Node.js:
 
 ```sh
 npx create-epic-app@latest


damiensedgwick:
Happy to close if it is felt that this is not needed. Only a tiny change so no effort on my part really.

kentcdodds (approved):
Thanks!

PR #983, opened 2025-04-03
Update CONTRIBUTING.md

kettanaito:
Fixes the contributing guidelines that imply that `prisma` CLI is globally installed. Uses `npx prisma` to use the same version of Prisma provided with Epic Stack. Generally, installing node modules globally is a bad idea and should be avoided.

## Checklist

- [ ] Tests updated
- [x] Docs updated


kentcdodds (approved):
Thanks!

PR #985, opened 2025-04-06
Move logo to components folder.

saurabhp75:
Move Logo component to components folder.

## Test Plan
- Run unit tests and e2e test to ensure there is no regression.

## Checklist

NA

### DIFF
--- a/app/components/logo.tsx
+++ b/app/components/logo.tsx
@@ -1,0 +1,14 @@
+import { Link } from "react-router";
+
+export function Logo() {
+	return (
+		<Link to="/" className="group grid leading-snug">
+			<span className="font-light transition group-hover:-translate-x-1">
+				epic
+			</span>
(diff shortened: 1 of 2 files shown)


kentcdodds (commented):
Why is this necessary?

saurabhp75:
Logo component looks out of place in root route, it should rather be in components folder where other pages can also import if required. The PR follows "full stack components" approach.

kentcdodds:
Thanks, but personally I don't think it's necessary to move it unless you really do need to use this component elsewhere. Thanks anyway!

saurabhp75:
Edit: Gives more flexibility to user in implementing the Logo and it's use anywhere. For eg. the user can add a "size" prop and use the logo in a form elsewhere.

Current implementation assumes that logo shall be used in root route, which imho is restricting.

Issue #998, opened 2025-04-23
Note editor contains errors of key prop being spread into JSX


justinvoelker:
When `__note-editor.tsx` renders, the console logs multiple errors.

```
A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, id: ..., name: ..., form: ..., type: ..., defaultValue: ...};
  <input {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {id: ..., name: ..., form: ..., type: ..., defaultValue: ...};
  <input key={someKey} {...props} />
```

An example of an offending input is `<input {...getInputProps(fields.id, { type: 'hidden' })} />`

One solution is to redeclare the key after the spread. A PR with the necessary changes is incoming.

PR #999, opened 2025-04-23
Fix key prop spread error

justinvoelker:
Resolves https://github.com/epicweb-dev/epic-stack/issues/998

## Test Plan

Checked UI console after making changes and the errors are no longer present

## Checklist

- [ ] Tests updated
- [ ] Docs updated

No tests or documentation updates necessary


kentcdodds (approved):
Thanks!

PR #1007, opened 2025-05-08
Improve vite development server config

AlemTuzlak:
There are a few things these PR aims to achieve:
1. Current setup runs the vite custom server in middleware mode but because we do not specify the appType it's set to SPA which runs HTML transforming middleware which we don't want to do as the "custom" property is the correct type to use (reference here: https://vite.dev/config/server-options.html#server-middlewaremode)
 

## Test Plan

Should work the same as before:
- local dev server starts normally and everything works

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds (approved):
Thanks!

PR #1009, opened 2025-05-09
Img alt should have the full name

alifhaider:
For e2e testing, it uses `name ?? username`

https://github.com/epicweb-dev/epic-stack/blob/b92c19ac1417ed13dac67f195560b8a03b8b5fb4/tests/e2e/settings-profile.test.ts#L59-L61


kentcdodds (approved):
Thanks!

PR #1012, opened 2025-05-22
upload source maps during build step

zwhitchcox:
Now that the image is built in the `Build Container` step instead of the `Deploy` step, we need to include the sentry auth token in the `Build Container` step. 

## Test Plan

I have tested this in production on our server. The source map is uploaded to sentry and referenced by a hash of the file, so you can upload as many source maps as you want, and they will not interfere with the source map in production.

### Before

Screenshot 2025-05-15 at 10 13 03 AM

### After

Screenshot 2025-05-22 at 3 37 12 PM


kentcdodds (approved):
Oh nice. Thank you! Good improvement on the deploy command too. Thanks!

PR #1013, opened 2025-05-27
Upgrade to latest openimg version

andrelandgraf:
openimg v1 auto-orients images based on EXIF tags. This fixes orientation issues with some images taken on iPhone and other modern smartphones. 

openimg v1 also provides a way to implement custom image pipelines with sharp, which is cool! :D

## Test Plan

N/A

## Checklist

- [X] Tests updated (N/A)
- [X] Docs updated (N/A)

## Screenshots

N/A


kentcdodds (approved):
Thanks!

PR #1015, opened 2025-06-09
replace bcrypt with scrypt

AdityaKirad:
refactor(auth): replace bcrypt with much better password hashing algorithm built into node crypto module for password hashing and validation
refactor(validation): update password schema to reflect scrypt limits
refactor(tests): use crypto for password hashing in test utilities

### DIFF
--- a/app/utils/auth.server.ts
+++ b/app/utils/auth.server.ts
@@ -319,7 +319,7 @@ async function generateKey(
 				N: SCRYPT_PARAMS.N,
 				r: SCRYPT_PARAMS.r,
 				p: SCRYPT_PARAMS.p,
-				maxmem: 128 * SCRYPT_PARAMS.N * SCRYPT_PARAMS.r * 2,
+				maxmem: 128 * SCRYPT_PARAMS.N * SCRYPT_PARAMS.r * SCRYPT_PARAMS.p * 2,
 			},
 			(err, key) => {
 				if (err) reject(err)
--- a/docs/decisions/045-scrypt-migration.md
+++ b/docs/decisions/045-scrypt-migration.md
@@ -43,11 +43,11 @@ These parameters were chosen to:
 - Maintain acceptable performance characteristics
 
 The actual scrypt options object includes an additional `maxmem` parameter set
-to `128 * N * r * 2`, which is approximately 64MiB for our parameters. This is
-explicitly set because Node.js has an internal default memory limit of 32 MiB.
-By setting this parameter, we're telling Node.js that twice the estimated memory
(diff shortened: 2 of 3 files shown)


kentcdodds:
I love built-in solutions. Thanks!

1. Could you write a decision doc? Include within it, a migration guide
2. Could you check why the tests are failing?

AdityaKirad:
bcrypt was specifically designed to cpu hard by increasing the cost we can increase the cpu it is resistant to the GPU brute force attack but it's not memory hard also we already know that in bcrypt we limitation over how many bytes password can have while scrypt is designed to memory hard with N,r,p parameters we can increase the cpu time and memory hardness also due to it's memory hardness it very hard for attackers using GPUs, ASICs also unlike the bcrypt it doesn't have any limit over the bytes in password basically scrypt makes very hard for attackers for cracking the password

AdityaKirad:
I forgot to mention how much memory `bcrypt` consumes since it's memory light it typically consumes around 4kb of memory per hash operation sometimes even less

kentcdodds (commented):
Thanks a lot for this! I'm a little bit concerned that this is going to cause issues for people running the Epic Stack within the free allowances. Can you show me what the Firecracker graphs look like for memory when passwords are being checked?

kentcdodds:
I'm thinking about Fly.io (which is where the Epic Stack is deployed). They have unofficial "forgiveness" of charges less than $5, https://fly.io/calculator

That gives you 256GB of memory. It's a tight budget for Node.js-based apps like ours. Most people have to bump that up pretty quick for real stuff, but I'd like simple projects/demos to be able to stay under that threshold.

AdityaKirad:
Hey Kent, I don't know much about the Firecracker Graph you are talking about but this something I recorded in the performance tab memory checkbox enabled during the login attempt with email-password and current scrypt config
image

andrecasal:
You can see the Firecracker graph here: https://fly.io/apps/<your-app-name>/metrics

There’s also a button on the top right of each card that takes you to Graphana.
IMG_0051

kentcdodds:
I'm not concerned about development. People have plenty of memory on their machines. I'm concerned about production.

AdityaKirad:
well in that case then you only can test the app in production as I don't have fly account, and I think the unofficial "forgiveness" of charges less than $5 only applies to the account which were created before the introduced pay as you go pricing

andreareginato:
This pull request is open since a while.

@AdityaKirad could you try to deploy an app with the current changes on Fly? It does not take much time and it would let us all understand if your solution can be integrated or not into the epic stack.

Thanks a lot

PR #1025, opened 2025-07-26
fix the existing note image replacement

wonu:
Fixes the incorrect usage of the `` component when `previewImage` was a data URL during existing note image replacement.

## Checklist

- [ ] Tests updated
- [ ] Docs updated


kentcdodds:
Thank you!

Issue #1029, opened 2025-08-09
Update imports decision after typescript update


jansepke:
In [031-imports.md](https://github.com/epicweb-dev/epic-stack/blob/main/docs/decisions/031-imports.md) it is mentioned that the epic stack needs typescript path aliases until [this PR](https://github.com/microsoft/TypeScript/pull/55015) is merged. This PR was released in Typescript 5.4 in March 2024 ([source](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#auto-import-support-for-subpath-imports)). So you can drop the path config fromt the tsconfig.json file and update the decision documentation.

kentcdodds:
PR welcome!

PR #1030, opened 2025-08-10
remove tsconfig aliases

jansepke:
as typescript reuses package.json import config

## Checklist

- [x] Docs updated

solves #1029

### DIFF
--- a/app/components/error-boundary.tsx
+++ b/app/components/error-boundary.tsx
@@ -6,7 +6,7 @@ import {
 	useParams,
 	useRouteError,
 } from 'react-router'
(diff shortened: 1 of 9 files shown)


jansepke:
a few imports had no or the wrong extension, please give it another try

kentcdodds:
This is great, thank you so much for working on it. Rather than updating an existing decision document, we like to keep those the same for historical reasons and instead create a new one that supersedes it. You can mark the current decision document as superseded and link to your new one. Could you do that?

kentcdodds:
Closed in favor of: #1054

PR #1034, opened 2025-08-25
fix: improve accessibility of the user menu

kettanaito:
When testing the Epic Stack app as a part of my recent workshop, I've noticed that certain UI elements have poor accessibility, which resulted in them being rather difficult to properly target during the tests. 

The user menu dropdown is one such example. Right now, it has an accessible name comprised of the username repeated twice:

Screenshot 2025-08-25 at 11 24 18

This happens because the dropdown has two elements, each of which has user's username as the accessible text:

- the avatar `img` as `alt={user.name}`
- the dropdown button content is `user.name`.

That results into the element's accessible test being "Kody Kody", which looks broken.

Instead, the element's accessible text should describe what that element is or does. In this case, I propose it should be something like "User menu".

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/components/user-dropdown.tsx
+++ b/app/components/user-dropdown.tsx
@@ -25,6 +25,7 @@ export function UserDropdown() {
 						// this is for progressive enhancement
 						onClick={(e) => e.preventDefault()}
 						className="flex items-center gap-2"
+						aria-label="User menu"
 					>
 						<Img
 							className="size-8 rounded-full object-cover"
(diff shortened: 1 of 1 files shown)


app/components/user-dropdown.tsx
```diff
@@ -25,13 +25,15 @@ export function UserDropdown() {
...
 						className="flex items-center gap-2"
+						aria-label="User menu"
 					>
 						<Img
 							className="size-8 rounded-full object-cover"
 							alt={user.name ?? user.username}
 							src={getUserImgSrc(user.image?.objectKey)}
 							width={256}
 							height={256}
+							aria-hidden="true"
```
kettanaito, line 36: The user avatar is a decorative element in this context and should be disregarded accessibility-wise. Even its `alt` is not needed, but I fear to anger eslint at this point.

kentcdodds (approved):
Yes, this is definitely better. Thanks!

PR #1035, opened 2025-08-25
fix: prevent user avatar from shrinking

kettanaito:
The user avatar element on the `/users/:username/notes` page _shrinks_, resulting in a broken UI:

Screenshot 2025-08-25 at 12 13 41

This is even more apparent with a placeholder avatar:

Screenshot 2025-07-14 at 10 25 21

This happens because the `Link` parent has `display: flex` and the text content (the user name) takes priority when it grows, shrinking the image.

Now, traditionally, you fix this by setting `flex-shrink: 0` on the element whose dimensions you wish to preserve. Alas, the component tree of the `Img` component doesn't make that fix possible as the classes you provide to the component end up on the `` element and not the parent `<picture />` element, which is the child of the flex container:

```
<Link>
  <picture> // <- shrink-0 needs to go here
    <img> // <- but it ends up going here
```

As such, I'm proposing the next best thing, which is to _fallback to the next responsive state of this element sooner_:

Screenshot 2025-08-25 at 12 18 33

This makes the avatar/username wrap sooner, preventing the shrunk avatar issue.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

### DIFF
--- a/app/routes/users+/$username_+/notes.tsx
+++ b/app/routes/users+/$username_+/notes.tsx
@@ -38,12 +38,12 @@ export default function NotesRoute({ loaderData }: Route.ComponentProps) {
 					<div className="absolute inset-0 flex flex-col">
 						<Link
 							to={`/users/${loaderData.owner.username}`}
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Good catch. Thanks!

Issue #1046, opened 2025-10-04
Note image duplication on navigate


damiensedgwick:
When navigating to a users notes, and navigating back and forth between the notes, sometimes, an image will duplicate multiple times. Refreshing the page resets the note images back to the correct amount.

This is on a fresh new project that was seeded and no code has been changed. I have not pushed the code up here so that it can be pulled / ran for verification because I feel like this might be a hard one to replicate and reproduce given the random nature of a seeded database. However I am happy to do so if needed.

https://github.com/user-attachments/assets/30e60ce9-ac9d-4fdc-91ef-34e77e1c147e

damiensedgwick:
> Thanks for the issue. Could you open a PR to fix it?

yeah sure!

damiensedgwick:
Thinking out loud here.

After a fair bit of testing and reproducing the issue, this only seems to be reproducible when there is a note that has duplicated images to begin with.

Take the following note where it has the same image twice, if I navigate to this note, then to another note, one of these images appear on the note I have navigated to even though it does not have a relationship with the note. Navigating back to the below note, the additional image persists in the UI. This can happen indefinitely but they can be cleared by refreshing the page.

```ts
{
  id: 'cmgfisaak001y8opph8gnvi7b',
  title: 'Colligo sortitus aliqua acsi subito curatio cernuus corporis thema.',
  content: 'Verbum infit abstergo optio doloribus officiis admoveo argentum adeo. Aduro ulterius dens comes abstergo vix caries adinventitias advenio. Vito sordeo defaeco nisi ustulo aperio.\n' +
    'Vacuus adflicto denique. Suasoria dolorum cresco aptus peior. Dolorem conforto vinco vigor sumptus confugo artificiose quae tamen terra.\n' +
    'Vespillo sono despecto censura pel comminor textilis at. Placeat nostrum tres carbo deleo alias tametsi tricesimus sumo. Circumvenio adeo articulus comburo adsuesco adimpleo convoco solvo adinventitias voluptas.',
  ownerId: 'cmgfisaah001o8oppk4xfvo0q',
  updatedAt: 2025-10-06T19:24:32.925Z,
  images: [
    {
      altText: 'an illustration of a hot air balloon',
      objectKey: 'notes/5.png'
    },
    {
      altText: 'an illustration of a hot air balloon',
      objectKey: 'notes/5.png'
    }
  ]
}
```

kentcdodds:
Thanks for the PR!

PR #1047, opened 2025-10-06
Fix note image duplication on navigate

damiensedgwick:
This merge request fixes an issue where images were being duplicated when navigating indefinitely. This _I think_ is down to how React reconciles elements that do not have unique keys.

The fix is to use the note image id for the map keys so that the uniqueness required, exists. I have also added additional code to ensure that when seeding the database, a note is not created using the same image multiple times. The second bit might be beyond the scope of this so I am happy to remove it, I've included it as a nice to have (in my opinion).

Tests have not been updated but no functionality has been lost although  I don't believe this area of the code is currently covered.

## Test Plan

A bit of a tedious one to test because a new project is needed with a seeded database where a note is created with the same image being used multiple times. Once you have this, you will be able to navigate back and forth and see the issue as shown in #1046

Pulling down my changes, you will see that this issue is no longer happening as we have swapped out using the objectKey for the map key in favour of the image id.

## Checklist

- [ ] Tests updated
- [ ] Docs updated

## Screenshots

Before:

https://github.com/user-attachments/assets/89683886-f56c-45f8-9a54-86853623b248

After:

https://github.com/user-attachments/assets/202144ca-dd27-4566-9fdd-503cf6251d92

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 node_modules
 .DS_store
 
+mise.toml
+
 /build
 /server-build
 .env
--- a/app/routes/users+/$username_+/notes.$noteId.tsx
+++ b/app/routes/users+/$username_+/notes.$noteId.tsx
@@ -32,6 +32,7 @@ export async function loader({ params }: Route.LoaderArgs) {
 			updatedAt: true,
 			images: {
 				select: {
+					id: true,
 					altText: true,
 					objectKey: true,
 				},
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,6 +1,8 @@
 node_modules
 .DS_store
 
+mise.toml
+
```
kentcdodds, line 5: I'd rather not include things in the gitignore for tools the stack does not officially use/recommend. Could this be in your global gitignore maybe?


```suggestion
```

kentcdodds (changes requested):
Nice 👍 Thanks!

.gitignore
damiensedgwick: oops my bad! sorry about that.

PR #1064, opened 2025-12-03
Update Prisma command to specific version 6

aerojeyenth:
Since npx points to the latest version of prisma which v7 which has breaking changes and this seed command fails  so pinning the npx version to 6 until we migrate the project to v7

## Test Plan

n/a

## Checklist

n/a

## Screenshots

n/a


kentcdodds (approved):
Thanks! Hopefully we get v7 going soon.

PR #1066, opened 2025-12-16
Migrate server code to `server/index.ts` and simplify processes

kasperpeulen:
When reviewing the code, I noticed custom build stuff that seemed unnecessary. It made me feel that react-router needs a lot of boilerplate, and custom stuff build scripts.

I looked at the best practices in react router:
https://github.com/remix-run/react-router-templates/tree/main/node-custom-server

And I concluded that the epic-stack does custom things that are not necessary. 

This gets rid of:
* The custom esbuild for the server files in other/build-server.ts
* The custom dev-server in server/dev-server.js
* The build:server command

For running the server files with ts without compilation or tsx, node 22.18 is needed, so I updated that requirement.

They could also easily be js files, as they contain very little types. That is what is done in the react-router-templates, but running TS through node for those 2 files, seems complety valid to me in modern code.

## Test Plan

build, dev, dev:no-mocks, start and start:mocks commands should still work the same way

I tested all of them manually.

## Checklist

- [x] Tests updated
- [x] Docs updated

### DIFF
--- a/app/routes/_auth/auth.$provider/callback.test.ts
+++ b/app/routes/_auth/auth.$provider/callback.test.ts
@@ -29,7 +29,7 @@ test('a new user goes to onboarding', async () => {
 	const response = await loader({
 		request,
 		params: PARAMS,
-		context: {} as AppLoadContext as AppLoadContext,
+		context: {} as AppLoadContext,
 	}).catch((e) => e)
 	expect(response).toHaveRedirect('/onboarding/github')
 })


kentcdodds (approved):
This is awesome! How is it done? I thought type stripping was only available without flags in node v25.

In any case, I absolutely love this if it works!

kentcdodds:
That's just fantastic. Thanks!

kasperpeulen:
Fixed here @davidesigner 
https://github.com/epicweb-dev/epic-stack/pull/1076

